PR #42, opened 2019-11-20
proTES TLS ingress support

djixyacine:
Add support for TLS termination for both the flower deployment
and proTES. Helm conditionals had to be used since the Ingress
objects definitions vary depending on the Ingress controller
which is on the target cluster. For OpenShift we use Route objects
and for Kubernetes we use Ingress objects.

The cluster type is defined under values.yaml, the possible values
are either 'openshift' or 'kubernetes'.

The documentation has been updated accordingly.

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -43,6 +43,10 @@ cd proTES
 vim values.yaml
 ```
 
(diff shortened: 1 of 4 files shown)


uniqueg:
Sounds great!

PR #54, opened 2020-05-06
Adding a basic version that supports JWT security

AlexanderSenf:
**Details**

setting `authorization_required: True` in `app_config.yaml` causes this error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/data/specs/20190903.d55bf88.task_execution_service.modified.swagger.security_definitions_added.yaml'
```
**Documentation**

This PR merely adds a Swagger file `20190903.d55bf88.task_execution_service.modified.swagger.security_definitions_added.yaml`
In this Swagger file JWT-based security is applied to all endpoints:
```
components:
  securitySchemes:
    bearerAuth:            # arbitrary name for the security scheme
      type: http
      scheme: bearer
      bearerFormat: JWT    # optional, arbitrary value for documentation purposes
security:                  # apply security globally to all endpoints
  - bearerAuth: []         # use the same name as above
```

No other changes were necessary. Now, if security is enabled, a JWT is expected and validated.

### DIFF
--- a/pro_tes/api/20190903.d55bf88.task_execution_service.modified.swagger.security_definitions_added.yaml
+++ b/pro_tes/api/20190903.d55bf88.task_execution_service.modified.swagger.security_definitions_added.yaml
@@ -1,0 +1,593 @@
+swagger: '3.0'
+info:
+  title: Task Execution Service
+  version: '0.4.1'
+components:
+  securitySchemes:
+    bearerAuth:            # arbitrary name for the security scheme
+      type: http
+      scheme: bearer
+      bearerFormat: JWT    # optional, arbitrary value for documentation purposes
+security:                  # apply security globally to all endpoints
+  - bearerAuth: []         # use the same name as above
+schemes:
+  - http
+consumes:
+  - application/json
+produces:
+  - application/json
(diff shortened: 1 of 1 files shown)


uniqueg:
Thanks a lot @AlexanderSenf!

In principle the function `__add_security_definitions` in <https://github.com/elixir-cloud-aai/proTES/blob/dev/pro_tes/api/register_openapi.py> should add the following security definitions to the specs and save them in file `20190903.d55bf88.task_execution_service.modified.swagger.security_definitions_added.yaml`:

```yaml
securityDefinitions:
  jwt:
    type: apiKey
    name: Authorization
    in: header
```

I'll see if I can reproduce the file not found and then try to fix that (must be something with the path, or perhaps `/data/specs` needs to be created first), because I think I'd like to avoid having two different specs for auth switched on/off, because it opens a door to the specs diverging in the future, for the two run modes. I'll try to do it until next week or so.

AlexanderSenf:
Thank you for the quick response @uniqueg. I agree with not wanting to have different specs, if possible!

uniqueg:
Closing in favor of #71 which should allow configuring auth via FOCA

PR #74, opened 2022-08-04
feat: track task progress asynchronously

Ayush5120:
To add track task progress in the background using celery and rabbitmq for endpoint `POST /tasks.


pro_tes/ga4gh/tes/models.py
```diff
@@ -335,26 +336,6 @@ class TesState(Enum):
...
-    @property
-    def is_cancelable(self):
-        """Check if a state is among the set of cancelable states."""
-        return self in (
-            self.QUEUED,
-            self.INITIALIZING,
-            self.RUNNING,
-            self.PAUSED,
-        )
-
```
uniqueg, line 357: I actually kinda preferred this over the definition of just the enums in `states.py`, but it's not a big deal, both work.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -515,43 +514,43 @@ def _sanitize_request(
...
+                        specify https://my.tes.com/api if the actual API is
+                        hosted at https://my.tes.com/api/ga4gh/tes/v1.
+                    remote_base_path: Override the default path suffix
+                        defined in the TES API specification,
+                        i.e., `/ga4gh/tes/v1`.
+                    remote_task_id: Task identifier on remote WES service.
+                    jwt: Authorization bearer token to be passed on with
+                         task request to external engine.
+                    timeout: Timeout for the job. Set to `None` to disable
+                             timeout.
```
uniqueg: This section shouldn't be indented that much. `Args:` should be the same indentation as the start of the docstring. Check other docstrings for comparison.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -515,43 +514,43 @@ def _sanitize_request(
...
-        #         'jwt': jwt,
-        #         'worker_id': worker_id,
-        #         'remote_host': remote_host,
-        #         'remote_base_path': remote_base_path,
-        #         'remote_task_id': remote_task_id,
-        #     },
-        #     soft_time_limit=timeout,
-        # )
-        # return None
+    def _track_progress(
```
uniqueg: It kinda doesn't make sense to define a separate function here, as it's doing only a single call. Just call it directly. Might be that I have it in proWES like that, didn't check, if so, perhaps you could fix it here and then raise a PR in proWES as well. It's really unnecessarily bloating the codebase and repeating function documentation.

pro_tes/tasks/tasks/track_task_progress.py
```diff
@@ -0,0 +1,91 @@
...
+from flask import (Flask, current_app)
+
+from pro_tes.ga4gh.tes.models import (
+    TesState
+)
+
+import tes
+from pro_tes.celery_worker import celery
+from pro_tes.ga4gh.tes.states import States
+logger = logging.getLogger(__name__)
```
uniqueg: Please mind the order of imports:

1. Imports from built-in modules
2. Imports from third party packages
3. Imports from _this_ package

Please order each group alphabetically by the name of the module/package. And please leave one empty line between each block.

Also, please leave one line between the imports and any following statements, like the logger initialization.

You can check other modules for comparison, as well as the official PEP 8 Python style guide and the PEP on docstrings. (Note that these don't prescribe ordering of modules/packages within each of the three blocks, but I find it improves readability and maintainability)

uniqueg (changes requested):
Thanks @Ayush5120! Looks good to me, just a couple minor issues, see comments

pro_tes/ga4gh/tes/task_runs.py

pro_tes/ga4gh/tes/task_runs.py

uniqueg (approved):

pro_tes/tasks/tasks/track_task_progress.py
uniqueg: Thanks. Imports from this module aren't actually sorted alphabetically by module, but okay, not a big problem. Just try to stick to it in the future. It's certainly helpful if you have a lot of imports.

PR #77, opened 2022-10-27
Minor fix

Ayush5120:
1. Added exceptions for tasks not found in  `GET /tasks/{id}` endpoint.
2. Removed unnecessary config files.


uniqueg:
Removing config handler is now addressed here: https://github.com/elixir-cloud-aai/proTES/pull/100

uniqueg (changes requested):
Additional errors look like a good thing, but it seems to me that (1) `EngineProblem` should also show up in the exceptions dictionary and (2) these errors should actually be used somewhere in the code?

Re: removal of config parser, see other message (basically, this is already taken care of in another PR)

Ayush5120:
> Additional errors look like a good thing, but it seems to me that (1) `EngineProblem` should also show up in the exceptions dictionary and (2) these errors should actually be used somewhere in the code?
> 
> Re: removal of config parser, see other message (basically, this is already taken care of in another PR)

Ok I will add `EngineProblem ` as stated and (2) Yeah I do found that most of the errors are defined and never used I will fix them in next commit.

uniqueg:
I'm not sure if this PR is still needed after refactoring proTES in November. Can you please check @Ayush5120, and either update or close the PR? :pray:

Ayush5120:
Yes @uniqueg you are right, this was handled in the pr [#114](https://github.com/elixir-cloud-aai/proTES/pull/114). Hence closing this PR.

PR #118, opened 2023-01-12
fix: store task updates in correct schema property

Ayush5120:
Fixed background task progress by storing the output of the task at the correct root which is "task_outgoing".

### DIFF
--- a/pro_tes/tasks/track_task_progress.py
+++ b/pro_tes/tasks/track_task_progress.py
@@ -2,9 +2,7 @@
 
 import logging
 from time import sleep
-from typing import (
-    Dict,
-)
+from typing import Dict
 
 from foca.database.register_mongodb import _create_mongo_client
 from foca.models.config import Config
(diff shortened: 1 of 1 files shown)


pro_tes/tasks/track_task_progress.py
```diff
@@ -2,36 +2,36 @@
...
-    self,  # pylint: disable=unused-argument
-    worker_id: str,
-    remote_host: str,
-    remote_base_path: str,
-    remote_task_id: str,
+        self,  # pylint: disable=unused-argument
+        worker_id: str,
+        remote_host: str,
+        remote_base_path: str,
+        remote_task_id: str,
```
uniqueg: Did `black` auto-format that in this way? I'm pretty sure there should be only one level of indentation relative to the function `def` here...

uniqueg (changes requested):
Looks good to me, thanks! Please double check the indentation. Specifically, please unindent the function args by one level and then run Black. Does it change again to two levels of indentation? If so, let me know and we'll just go with that. If it keeps the one level indentation, please commit and push and also let me know. Rest is fine, so I'll merge as soon as I hear from you, one way or the other.

pro_tes/tasks/track_task_progress.py
Ayush5120: Yes, you are right, I used black it reformatted to 1 level of indentation. fixing in the next commit

uniqueg: Great, thanks

uniqueg (approved):

PR #119, opened 2023-01-19
feat: distance-based task distribution logic

Ayush5120:
Added task distribution logic, that computes the total distance between the set of input URIs (from the incoming TES task) and the available TES instances, then relays the task to the closest TES instance.

### DIFF
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -30,6 +30,7 @@ from pro_tes.utils.db import DbDocumentConnector
 from pro_tes.utils.models import TaskModelConverter
 
 # pragma pylint: disable=invalid-name,redefined-builtin,unused-argument
+# pragma pylint: disable=too-many-locals
 
 logger = logging.getLogger(__name__)
 
@@ -63,8 +64,7 @@ class TaskRuns:
         payload: Dict = deepcopy(request.json)
 
         db_document: DbDocument = DbDocument()
-        db_document.tes_endpoint = TesEndpoint(host=payload["tes_uri"])
-
+        tes_uri_list = deepcopy(payload["tes_uri"])
         del payload["tes_uri"]
 
         db_document.task_incoming = TesTask(**payload)
@@ -101,63 +101,90 @@ class TaskRuns:
                 f"Original error message: '{type(exc).__name__}: "
                 f"{exc}'"
             ) from exc
+
         try:
-            cli = tes.HTTPClient(url, timeout=5)
-        except ValueError as exc:
+            for tes_uri in tes_uri_list:
+                db_document.tes_endpoint = TesEndpoint(host=tes_uri)
+                url: str = (
+                    f"{db_document.tes_endpoint.host.rstrip('/')}/"
+                    f"{db_document.tes_endpoint.base_path.lstrip('/')}"
+                )
+                try:
+                    cli = tes.HTTPClient(url, timeout=5)
+                except ValueError as exc:
+                    db_connector.update_task_state(
+                        state=TesState.SYSTEM_ERROR.value
+                    )
+                    logger.info(
+                        f"Task '{db_document.task_incoming.id}' could not "
+                        f"be sentto TES endpoint hosted at: {url}. Invalid TES"
+                        f" endpoint URL. Original error message: "
+                        f"'{type(exc).__name__}: {exc}'"
+                    )
+                    continue
+
+                try:
+                    task_id = cli.create_task(payload_marshalled)
+                except requests.HTTPError as exc:
+                    db_connector.update_task_state(
+                        state=TesState.SYSTEM_ERROR.value
+                    )
+                    logger.info(
+                        f"Task '{db_document.task_incoming.id}' "
+                        f"could not be sent to TES endpoint hosted "
+                        f"at: {url}. Task could not be created. Original "
+                        f"error message: '{type(exc).__name__}: "
+                        f"{exc}'"
+                    )
+                    continue
+
+                logger.info(
+                    f"Task '{db_document.task_incoming.id}' "
+                    f"forwarded to TES endpoint "
+                    f"hosted at: {url}. proTES task identifier: {task_id}."
+                )
+                try:
+                    task: Task = cli.get_task(task_id)
+                    task_model_converter = TaskModelConverter(task=task)
+                    task_converted: TesTask = \
+                        task_model_converter.convert_task()
+                    db_document = db_connector.upsert_fields_in_root_object(
+                        root="task_outgoing",
+                        **task_converted.dict(),
+                    )
+                except requests.HTTPError as exc:
+                    logger.error(
+                        f"Task '{db_document.task_incoming.id}' info could "
+                        f"not be retrieved from TES endpoint hosted at: "
+                        f"{url}. Original error message: "
+                        f"'{type(exc).__name__}: {exc}'"
+                    )
+                except PyMongoError as exc:
+                    logger.error(
+                        "Database could not be updated with task info "
+                        f"retrieved for task '{db_document.task_incoming.id}'"
+                        f"sent to TES endpoint hosted at: {url}. "
+                        f"Original error message:'{type(exc).__name__}: {exc}'"
+                    )
+                task__track_task_progress.apply_async(
+                    None,
+                    {
+                        "worker_id": db_document.worker_id,
+                        "remote_host": db_document.tes_endpoint.host,
+                        "remote_base_path": db_document.tes_endpoint.base_path,
+                        "remote_task_id": db_document.task_outgoing.id,
+                    },
+                )
+                return {"id": task_id}
+
+        except Exception as exc:
             db_connector.update_task_state(state=TesState.SYSTEM_ERROR.value)
             raise InternalServerError(
                 f"Task '{db_document.task_incoming.id}' could not be sent "
-                f"to TES endpoint hosted at: {url}. Invalid TES endpoint URL. "
+                f"to any of the TES endpoint available"
                 f"Original error message: '{type(exc).__name__}: "
                 f"{exc}'"
             ) from exc
-        try:
-            task_id = cli.create_task(payload_marshalled)
-        except requests.HTTPError as exc:
-            db_connector.update_task_state(state=TesState.SYSTEM_ERROR.value)
-            raise InternalServerError(
-                f"Task '{db_document.task_incoming.id}' could not be sent "
-                f"to TES endpoint hosted at: {url}. Task could not be "
-                f"created. Original error message: '{type(exc).__name__}: "
-                f"{exc}'"
-            ) from exc
-        logger.info(
-            f"Task '{db_document.task_incoming.id}' forwarded to TES endpoint "
-            f"hosted at: {url}. proTES task identifier: {task_id}."
-        )
-        try:
-            task: Task = cli.get_task(task_id)
-            task_model_converter = TaskModelConverter(task=task)
-            task_converted: TesTask = task_model_converter.convert_task()
-            db_document = db_connector.upsert_fields_in_root_object(
-                root="task_outgoing",
-                **task_converted.dict(),
-            )
-        except requests.HTTPError as exc:
-            logger.error(
-                f"Task '{db_document.task_incoming.id}' info could not be "
-                f"retrieved from TES endpoint hosted at: {url}. Original "
-                f"error message: '{type(exc).__name__}: {exc}'"
-            )
-        except PyMongoError as exc:
-            logger.error(
-                "Database could not be updated with task info retrieved for "
-                f"task '{db_document.task_incoming.id}' sent to TES endpoint "
-                f"hosted at: {url}. Original error message: "
-                f"'{type(exc).__name__}: {exc}'"
-            )
-
-        task__track_task_progress.apply_async(
-            None,
-            {
-                "worker_id": db_document.worker_id,
-                "remote_host": db_document.tes_endpoint.host,
-                "remote_base_path": db_document.tes_endpoint.base_path,
-                "remote_task_id": db_document.task_outgoing.id,
-            },
-        )
-
-        return {"id": task_id}
 
     def list_tasks(self, **kwargs) -> Dict:
         """Return list of tasks.
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -1,10 +1,12 @@
 """Class implementing TES API-server-side controller methods."""
 
+import logging
 from copy import deepcopy
 from datetime import datetime
-import logging
 from typing import Dict, Tuple
 
+import requests
+import tes
 from bson.objectid import ObjectId
 from celery import uuid
 from dateutil.parser import parse as parse_time
@@ -13,17 +15,10 @@ from foca.models.config import Config
 from foca.utils.misc import generate_id
 from pymongo.collection import Collection
 from pymongo.errors import DuplicateKeyError, PyMongoError
-import requests
-import tes
 from tes.models import Task
 
 from pro_tes.exceptions import BadRequest, TaskNotFound
-from pro_tes.ga4gh.tes.models import (
-    DbDocument,
-    TesEndpoint,
-    TesState,
-    TesTask,
-)
+from pro_tes.ga4gh.tes.models import DbDocument, TesEndpoint, TesState, TesTask
 from pro_tes.ga4gh.tes.states import States
 from pro_tes.tasks.track_task_progress import task__track_task_progress
 from pro_tes.utils.db import DbDocumentConnector
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -17,7 +17,7 @@ import requests
 import tes
 from tes.models import Task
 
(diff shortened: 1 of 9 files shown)


pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
+"""Model for the Access Uri Combination."""
+# pylint: disable=no-name-in-module
+from typing import List
+from pydantic import BaseModel
+
+
+# pragma pylint: disable=too-few-public-methods
+class TesUriList(BaseModel):
```
uniqueg: This is not a list, but a single instance. Also, the model does not only include the URI, so it's perhaps better to rename it to `TesDeployment`. Finally, the current design is not great for extensibility if we produce/consider any other stats in the future. It might be better to introduce another layer of nesting and use an attribute `stats` here, an instance of model `TesStats` - which would currently contain a single attribute `total_distance`.

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+    total_distance: int = None
+
+
+class AccessUriCombination(BaseModel):
+    """Combination of input_uri of the TES task and the TES instances."""
+
+    input_uri: List[str]
+    tes_uri_list: List[TesUriList]
+
+    def convert_tes_uri_list_to_dict(self):
```
uniqueg: Why not simply use `pydantic`s `.dict()` method. Or if that does not do what you want (but it probably does), it's better to override it.

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+        """Convert the list of type TESUriList to list of dictionary."""
+        converted_list = []
+        for combo in self.tes_uri_list:
```
uniqueg: Could use list comprehension here to make this a one-liner

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+
+    tes_uri: str
+    total_distance: int = None
+
+
+class AccessUriCombination(BaseModel):
+    """Combination of input_uri of the TES task and the TES instances."""
+
+    input_uri: List[str]
+    tes_uri_list: List[TesUriList]
```
uniqueg: Better rename to `tes_deployments`, see above.

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+    """Combination of the tes_uri and total distance."""
+
+    tes_uri: str
+    total_distance: int = None
+
+
+class AccessUriCombination(BaseModel):
+    """Combination of input_uri of the TES task and the TES instances."""
+
+    input_uri: List[str]
```
uniqueg: Not sure if great for extensibility, perhaps we will consider other params in the future. So might be better to have another layer of nesting, as above. I would propose to name this `task_params`, an instance of `TaskParams`, with a single attribute `input_uris` (plural!), which is a list of URLs (not `str`) - check how you can do that in Pydantic [here](https://docs.pydantic.dev/usage/types/#urls).

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+# pylint: disable=no-name-in-module
+from typing import List
+from pydantic import BaseModel
+
+
+# pragma pylint: disable=too-few-public-methods
+class TesUriList(BaseModel):
+    """Combination of the tes_uri and total distance."""
+
+    tes_uri: str
```
uniqueg: Pydantic support various URL types: https://docs.pydantic.dev/usage/types/#urls
Better to use any of those

pro_tes/middleware/task_distribution.py
```diff
@@ -2,13 +2,22 @@
 
 from copy import deepcopy
```
uniqueg: I think it's better to have different modules for different task distribution logic implementations in a package `task_distribution`, i.e., random task distribution should go in `middleware/task_distribution/random.py` and task distribution by distance should go in `middleware/task_distribution/distance.py`.

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
+"""Model for the Access Uri Combination."""
+# pylint: disable=no-name-in-module
```
uniqueg: Put the pragma below the imports with the other pragma (btw, the `pragma` keyword is missing, so not sure this instruction will even work), and leave one line free between module docstring and imports. Also make sure to have different import sections, separated by empty lines (following [PEP 8](https://peps.python.org/pep-0008/#imports)):

1. Standard imports
2. Third party imports
3. Local application

Also, within each group, please order imports alphabetically by the name of the module you import or import from.

Please remember for the future.

pro_tes/middleware/models.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import List
+from pydantic import BaseModel
+
+
+# pragma pylint: disable=too-few-public-methods
```
uniqueg, line 11: Please leave one empty line between the imports and the pragmas, then two empty lines between the pragmas and the class definition.

pro_tes/middleware/models.py
uniqueg: Come to think of it, I would not use this pragma module-wide; just use it for the import where you need it

pro_tes/middleware/task_distribution.py
```diff
@@ -2,13 +2,22 @@
...
 import random
-from typing import List, Optional
+from urllib.parse import urlparse
+from socket import gaierror, gethostbyname
+from typing import Dict, List, Set, Tuple, Optional
+from itertools import combinations
+from ip2geotools.errors import InvalidRequestError
+from ip2geotools.databases.noncommercial import DbIpCity
+from geopy.distance import geodesic
```
uniqueg: Follow the import rules from PEP8 and use alphabetic sorting as described above. Check the other modules you changed as well.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -17,7 +17,7 @@
 import tes
```
uniqueg, line 17: It's difficult to parse the changes. Can you explain what you changed in this module and why? It doesn't seem obvious to me that this module needed to be changed, or at least not that extensively.

pro_tes/middleware/task_distribution.py
```diff
@@ -2,13 +2,22 @@
...
 
 from flask import current_app
 import requests
 
+from pro_tes.middleware.models import AccessUriCombination, TesUriList
 
-def task_distribution() -> Optional[str]:
+
+# pylint: disable-msg=too-many-locals
+def random_task_distribution() -> Optional[str]:
```
uniqueg: Both task distribution logic models should have the same signature, but here they return different things (str, List). I think what they should return is a (ranked) list of URLs (not strings).

I am also not sure whether `None` should be acceptable (rather return an empty list)? If there is a specific meaning of `None`, please explain.

pro_tes/middleware/middleware.py
```diff
@@ -27,10 +29,23 @@ class TaskDistributionMiddleware(AbstractMiddleware):
 
     def __init__(self):
         """Construct object instance."""
-        self.tes_uri = task_distribution()
+        self.tes_uri = []
+        self.input_uri = []
 
     def modify_request(self, request) -> Dict:
-        """Add TES instance to request body."""
+        """Add the best possible TES instance to request body."""
```
uniqueg, line 32: I think it would be best if the user/admin could change the task distribution logic to be used (currently `random` or `distance`). So please add a config param to `config.yaml` that points to the module to be used for task distribution (e.g., `pro_tes.middleware.task_distribution.distance`). This would then also allow using a task distribution logic from an external package.

We will need this setting so that we can compare the efficiency of different task distribution models.

uniqueg (changes requested):
Thanks @Ayush5120. There are quite some comments I made. Please address those and let's have another round of review after.

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Previously, we were trying to submit the task on the single provided TES instance, while now we have a list of ranked TES instances so we have to try for each in order,
Things that I have changed are:
1. Placed the submission of task code inside a for loop, where it tries to submit the task on the instance in the list at the 0th index, otherwise retries with the next instances.
2. Updating the TES instance in the database where the task got submitted.

pro_tes/middleware/models.py
Ayush5120: Understood. will fix in next commit.

pro_tes/middleware/models.py
Ayush5120: Ok I will try and fix it

pro_tes/ga4gh/tes/task_runs.py
uniqueg: Thanks, that's fine. Was just wondering why there were so many changes, but guess it's because of changes in formatting.

I'm not too happy with catching multiple errors across different calls, but this isn't something you introduced here, so let's leave it like that :)

pro_tes/middleware/__init__.py
```diff
@@ -1 +1 @@
-"""proTES middlwares."""
+"""proTES middlewares."""
```
uniqueg: Middleware is generally uncountable, so best not to use in plural.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -1,10 +1,12 @@
 """Class implementing TES API-server-side controller methods."""
 
+import logging
 from copy import deepcopy
 from datetime import datetime
-import logging
```
uniqueg: Not sure why you changed the order here, it was fine as it was - alphabetically _by module_, not by keyword.

That is:

```
import module1  # comes first because module1 < module2, even though import > from
from module2 import something
```

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -1,10 +1,12 @@
...
 from typing import Dict, Tuple
 
+import requests
```
uniqueg: Same - the sort order of `requests` and `tes` was fine

pro_tes/middleware/middleware.py
```diff
@@ -1,17 +1,18 @@
...
-    Dict,
-)
+from typing import Dict
 
-from pro_tes.middleware.task_distribution import (
+from pro_tes.middleware.task_distribution.distance import (
     task_distribution_by_distance,
+)
+from pro_tes.middleware.task_distribution.random import (
     random_task_distribution,
```
uniqueg: Best to have consistent naming patterns. Here, you call one `task_distribution_by_distance` and the other one `random_task_distribution`. Better to use the following names:
* `task_distribution_distance`
* `task_distribution_random`

This makes it much easier to understand that these are both implementations of an abstract concept.

You could even name both of them just `task_distribution`. To avoid conflicts, then, what you could do is:

```python
from pro_tes.middleware.task_distribution import distance, random

# use like this:
random.task_distribution()
distance.task_distribution()
```

pro_tes/middleware/models.py
```diff
@@ -1,31 +1,35 @@
...
-# pylint: disable=no-name-in-module
+
+# pragma pylint: disable=no-name-in-module
+
 from typing import List
-from pydantic import BaseModel
 
+from pydantic import AnyUrl, BaseModel, HttpUrl
 
 # pragma pylint: disable=too-few-public-methods
```
uniqueg: As I said before: don't use module-level pragmas unless you have a good reason, e.g., disabling `too-few-public-methods` totally makes sense in a module full of models. But it's not clear to me why `no-name-in-module` should be disabled in every line of the module. Unless I'm missing something, please disable the linter rule in line, wherever you need it.

If I _am_ missing something, and you really need to disable this in more than two places (and possibly more in the future), then please merge it with the other pragma further below.

Module organization should always be:
- 1st: Module-level docstring
- 2nd: Imports
- Everything else

pro_tes/middleware/models.py
```diff
@@ -1,31 +1,35 @@
...
+
+    tes_uri: HttpUrl
+    stats: TesStats
+
+
+class AccessUriCombination(BaseModel):
+    """Combination of input_uri of the TES task and the TES instances."""
+
+    task_params: TaskParams
+    tes_deployments: List[TesDeployment]
```
uniqueg: Good stuff :)

pro_tes/middleware/task_distribution/__init__.py
```diff
@@ -0,0 +1 @@
+"""proTES task distribution middlewares."""
```
uniqueg: Again, middleware (singular) instead of middlwares (plural)

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -1,42 +1,25 @@
-"""Module for task distribution logic."""
+"""Module for distance based task distribution logic ."""
```
uniqueg: "distance-based" with hyphen.

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -1,42 +1,25 @@
...
+from typing import Dict, List, Optional, Set, Tuple
+from urllib.parse import urlparse
 
 from flask import current_app
-import requests
-
-from pro_tes.middleware.models import AccessUriCombination, TesUriList
+from geopy.distance import geodesic
+from ip2geotools.databases.noncommercial import DbIpCity
+from ip2geotools.errors import InvalidRequestError
```
uniqueg, line 12: Those should be in the second block, not the third. The third is only for imports from the `pro_tes` package (absolute or relative).

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -1,42 +1,25 @@
...
+from pro_tes.middleware.models import (
+    AccessUriCombination,
+    TaskParams,
+    TesDeployment,
+    TesStats
+)
 
+# pylint: disable-msg=R0912
 # pylint: disable-msg=too-many-locals
-def random_task_distribution() -> Optional[str]:
```
uniqueg: Please use expressive Pylint rules (`too-many-branches` better than `R0912`).

uniqueg: By the way, rules such as `too-many-branches` and `too-many-locals` are sensible rules. They may be cases where it makes sense to disable, e.g., if you have a clean, specialized function that goes through lots of conditions. But in _most_ cases, I think what they should suggest to you to carefully check your code for smells:
* Is every local _really_ needed?
* Is every branch _really_ needed?
* If both answers are yes: Could the function perhaps be refactored into multiple subfunctions? Or perhaps a class-based structure?

uniqueg (changes requested):
Great - we're getting there! :)
I think all the remaining issues are small fry that shouldn't take long to address

pro_tes/middleware/task_distribution/distance.py
Ayush5120: This is already in the second block.

pro_tes/middleware/task_distribution/distance.py
Ayush5120: Understood will try to eliminate these errors rather than disabling them.

pro_tes/middleware/task_distribution/distance.py
uniqueg: Oh okay, probably I was thrown off by the red and green. Great then :)

pro_tes/middleware/models.py
Ayush5120: Actually, the error was arising from one of the import lines which is `from pydantic import AnyUrl, BaseModel, HttpUrl` that's why I have used `no-name-in-module` before imports. Maybe I should disable this in the same line as you mentioned above.

pro_tes/middleware/middleware.py
Ayush5120: okay, will fix in next commit

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: I was using [isort ](https://pycqa.github.io/isort/) for sorting the imports, if you want it the way it was before I will fix them in the next commit.

pro_tes/middleware/middleware.py
uniqueg: Okay, let's do this in a separate feature, fair enough. I'll open an issue pointing to this thread.

uniqueg (approved):
Thanks a lot @Ayush5120. I didn't try it out, really, but it looks good to me and I will merge it. I hope edge cases (e.g., no inputs at all) will work. But if not, we can fix them after merging. Note that I have opened #124 to handle middleware more gracefully and generically. It's not an urgent issue, though - we can take care of it after #123 and #122.

pro_tes/ga4gh/tes/task_runs.py
uniqueg: Ah okay. Fine to use `isort`, but the important thing is consistency. So either use the currently used style in your proposed changes OR change it according to `isort` conventions in _all_ modules (but then that should be done in a separate PR). Anyway, leaving this now, it's not _that_ important.

uniqueg:
Hi @Ayush5120. Thanks a lot for the great work :)

I am merging now, as I guess you have addressed at least most of the comments (and I have opened #124). If there are any more issues popping up during testing, we can open individual issues for them and fix them :)

In the future, I think it's best if you resolve open conversations yourself if you feel confident that you have addressed them non-controversially. Otherwise, please leave a comment, either describing how you addressed it, or why you did _not_ address it. That makes reviewing a whole lot easier.

Issue #123, opened 2023-01-26
Return execution trace in task log


uniqueg:
Currently, there is no way for clients to trace the route the task took from proTES, and so they have no chance of knowing where a given task was actually executed. This may or may not be desirable.

Implement a config param that when set to `True` adds information to the log on the trace that a given task request took. The information should extend the `tesTaskLog` model and account for the possibility that multiple gateways may be included in a call chain.

One solution might be the following recursive extension of `tesTaskLog`:

```yaml
    tesTaskLog:
      ...
      properties:
        ...
        forwarded_to:
          $ref: '#/components/schemas/tesNextTes'
      description: TaskLog describes logging information related to a Task.
```

with:

```yaml
    tesNextTes:
      required:
      - url
      - id
      type: object
      properties:
        url:
          type: string
          description: TES server to which the task was forwarded.
          example: https://my.tes.instance/
        id:
          type: string
          description: Task identifier assigned by the TES server to which the task was forwarded.
          example: job-0012345
        forwarded_to:
          $ref: '#/components/schemas/tesNextTes'
      description: Describes the TES server to which the task was forwarded, if applicable.
```

PR #128, opened 2023-01-31
fix: return internal task ID & refactor db model

### LINKED ISSUE #122: Return proTES task ID rather than external task ID
Currently `POST`ing a task returns the _external_ task ID, i.e., the task ID of the TES service that the task was forwarded to. However, proTES also mints its own internal task ID for each incoming task. It is this ID that should be returned to the client, and which should be used to `GET` tasks and listed in the `GET /tasks` response.

Ayush5120:
1. Fixed #122 
2. Now the proTES task id is returned instead of the remote task id ( Actual TES id where the task is executed).
3. Modified task_incoming logs in such a way that logs (that is, outputs and TES Executor logs) are updated from Actual TES, and the rest of the things (start_time, end_time, etc) are updated from the proTES.
4. Currently the information about the TES URI (where the task is submitted) and remote task id is stored in task incoming logs ( inside metadata), it is not the best way but it was essential to do this now because the endpoint `POST /tasks/{id}:cancel` relies on remote task id, this could be further fixed in #123

### DIFF
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -28,6 +28,7 @@ from pro_tes.utils.models import TaskModelConverter
 
 # pragma pylint: disable=invalid-name,redefined-builtin,unused-argument
 # pragma pylint: disable=too-many-locals
+# pylint: disable=unsubscriptable-object
 
 logger = logging.getLogger(__name__)
 
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -18,7 +18,9 @@ import tes
 from tes.models import Task
 
 from pro_tes.exceptions import BadRequest, TaskNotFound
-from pro_tes.ga4gh.tes.models import DbDocument, TesEndpoint, TesState, TesTask
+from pro_tes.ga4gh.tes.models import (
+    DbDocument, TesEndpoint, TesState, TesTask, TesTaskLog
+)
 from pro_tes.ga4gh.tes.states import States
 from pro_tes.tasks.track_task_progress import task__track_task_progress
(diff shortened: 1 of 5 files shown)


pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -410,3 +401,80 @@ def _set_projection(self, view: str) -> Dict:
...
+    ) -> DbDocument:
+        """Update the task incoming object."""
+        logs = self._set_logs(
+            payloads=deepcopy(payload),
+            start_time=kwargs["start_time"]
+        )
+        db_document.task_incoming.logs = [
+            TesTaskLog(**logs) for logs in logs
+        ]
+        db_document.task_incoming.state = TesState.UNKNOWN
```
uniqueg, line 419: I'm not feeling well and it's not easy to process these changes for me right now. But this seems odd to me. The task state in the incoming task model instance (which we use to return responses to the client) does of course need to be updated from the remote, just like the logs. It shouldn't just remain in state `UNKNOWN`.

But maybe I'm missing something here, it's really hard to concentrate.

Ayush5120: Well, the state is set to UNKNOWN in the initial setting, while it will be again updated once the task is submitted and it will change from `UNKNOWN` to `QUEUE->RUNNING->COMPLETED` once the background async tracking of the task starts.

uniqueg: Okay, fine then :) I'll just merge then, so that you can go on. We can debug/refactor later, if needed.

Ayush5120: The same goes for the TESexecutor Logs (i.e, logs inside logs ), and logs outputs will update asynchronously, while the other things in logs will remain the same as set in here.

PR #129, opened 2023-02-02
fix: return execution trace in task log

### LINKED ISSUE #123: Return execution trace in task log

Ayush5120:
- Fixed #123 
- Updated schema by nesting the TesNextTes inside the metadata of the TesTaskLog.
- Added param in the config file if set to true the remote_tes_id and TES URL (where the task is executed) is stored in log metadata to trace the execution of the task.

### DIFF
--- a/pro_tes/api/additional_logs.yaml
+++ b/pro_tes/api/additional_logs.yaml
@@ -1,0 +1,26 @@
+components:
+  schemas:
+    tesTaskLog:
+        properties:
+          metadata:
+            properties:
+              forwarded_to:
+                $ref: '#/components/schemas/tesNextTes'
+                description: TaskLog describes logging information related to a Task.
+    tesNextTes:
+      required:
+        - url
+        - id
+      type: object
+      properties:
+        url:
+          type: string
+          description: TES server to which the task was forwarded.
+          example: https://my.tes.instance/
+        id:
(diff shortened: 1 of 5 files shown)


uniqueg (approved):
LGTM

PR #133, opened 2023-02-07
fix: call middleware in task creation class

### LINKED ISSUE #132: Middleware called before task is created
The middleware is called before the task is even created, which results in a modified incoming task request and we can never store the unmodified copy of the original request
The ideal way is that the incoming task request is stored first in the database and then the middleware is called

Ayush5120:
- fix #132 by calling middleware in `pro_tes.ga4gh.tes.task_runs.TaskRuns` class rather than the upstream `pro_tes.ga4gh.tes.server.CreateTask()` controller
- the incoming request is now stored in the database before being modified by any middleware

### DIFF
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -71,9 +71,9 @@ class TaskRuns:
         Returns:
             Task identifier.
         """
+        start_time = datetime.now().strftime("%m-%d-%Y %H:%M:%S")
         payload: Dict = deepcopy(request.json)
         db_document: DbDocument = DbDocument()
-        start_time = datetime.now().strftime("%m-%d-%Y %H:%M:%S")
 
         db_document.basic_auth = self.parse_basic_auth(request.authorization)
 
@@ -114,7 +114,6 @@ class TaskRuns:
             ) from exc
 
         for tes_uri in tes_uri_list:
-
             db_document.tes_endpoint = TesEndpoint(host=tes_uri)
             url: str = (
                 f"{db_document.tes_endpoint.host.rstrip('/')}/"
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -66,8 +66,7 @@ class TaskRuns:
         """Start task.
 
         Args:
-            **kwargs: Additional keyword arguments passed along with
-                             request.
+            **kwargs: Additional keyword arguments passed along with request.
         Returns:
             Task identifier.
         """
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -29,6 +29,7 @@ from pro_tes.ga4gh.tes.models import (
     TesNextTes,
 )
 from pro_tes.ga4gh.tes.states import States
+from pro_tes.middleware.middleware import TaskDistributionMiddleware
 from pro_tes.tasks.track_task_progress import task__track_task_progress
 from pro_tes.utils.db import DbDocumentConnector
 from pro_tes.utils.models import TaskModelConverter
@@ -57,6 +58,7 @@ class TaskRuns:
             self.foca_config.db.dbs["taskStore"].collections["tasks"].client
         )
         self.store_logs = self.foca_config.storeLogs["execution_trace"]
+        self.task_distributor = TaskDistributionMiddleware()
 
     def create_task(  # pylint: disable=too-many-statements,too-many-branches
         self, **kwargs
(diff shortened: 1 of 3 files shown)


pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -67,12 +67,13 @@ def create_task(  # pylint: disable=too-many-statements,too-many-branches
 
         Args:
             **kwargs: Additional keyword arguments passed along with
-                             request and start_time.
+                             request.
```
uniqueg: What's with the weird indentation? I know you haven't added it in this comment, but you might as well fix it. Usually, indentation in Google-style docstrings for functions & methods is like this:

```console
"""1-line summary; start with verb in imperative & don't exceed 80 chars

Optional extended description, wrapping when max line length of 80 chars
is exceeded. No indentation on successive lines.

Args: (same for "Attributes" & "Raises" sections)
    keyword: keyword is indented one level
    keyword_2: if the keyword description exceeds 80 chars, then
        indent any additional lines by _one_ more level.

Returns:
    Summary description of return value in one line.

    If more info is needed, add it to an additional line that is separated
    by one empty line. Just like in the extended function/method
    description, do not use extra indentation.

"""
```

[Here is more info](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html), including another detailed example:

```python
def module_level_function(param1, param2=None, *args, **kwargs):
    """This is an example of a module level function.

    Function parameters should be documented in the ``Args`` section. The name
    of each parameter is required. The type and description of each parameter
    is optional, but should be included if not obvious.

    If \*args or \*\*kwargs are accepted,
    they should be listed as ``*args`` and ``**kwargs``.

    The format for a parameter is::

        name (type): description
            The description may span multiple lines. Following
            lines should be indented. The "(type)" is optional.

            Multiple paragraphs are supported in parameter
            descriptions.

    Args:
        param1 (int): The first parameter.
        param2 (:obj:`str`, optional): The second parameter. Defaults to None.
            Second line of description should be indented.
        *args: Variable length argument list.
        **kwargs: Arbitrary keyword arguments.

    Returns:
        bool: True if successful, False otherwise.

        The return type is optional and may be specified at the beginning of
        the ``Returns`` section followed by a colon.

        The ``Returns`` section may span multiple lines and paragraphs.
        Following lines should be indented to match the first line.

        The ``Returns`` section supports any reStructuredText formatting,
        including literal blocks::

            {
                'param1': param1,
                'param2': param2
            }

    Raises:
        AttributeError: The ``Raises`` section is a list of all exceptions
            that are relevant to the interface.
        ValueError: If `param2` is equal to `param1`.

    """
```

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -67,12 +67,13 @@ def create_task(  # pylint: disable=too-many-statements,too-many-branches
...
         Returns:
             Task identifier.
         """
         payload: Dict = deepcopy(request.json)
         db_document: DbDocument = DbDocument()
+        start_time = datetime.now().strftime("%m-%d-%Y %H:%M:%S")
```
uniqueg: Put this even higher. It should be the first call in the method.

uniqueg (changes requested):
Thanks!

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: This is fixed already in the latter commits, (my mistake I just addressed this change with the wrong commit message that was 'fixed linting').

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Will fix this indentation in the next commit, 
Moreover, there are issues in the explanation of the docstring of multiple functions in the codebase will create a separate issue for that and fix it there only.

uniqueg: That makes sense, thanks a lot :)

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -57,30 +58,39 @@ def __init__(self) -> None:
...
+        db_document.task_outgoing = TesTask(**payload)
+
+        # middleware is called after the task is created in the database
+        payload = self.task_distributor.modify_request(request=request).json
+
         tes_uri_list = deepcopy(payload["tes_uri"])
         del payload["tes_uri"]
 
-        db_document.task_outgoing = TesTask(**payload)
         db_document.task_incoming = TesTask(**payload)
```
uniqueg, line 88: I think I'm really confused. Didn't we say that we do _not_ want to modify the incoming but only the outgoing request object? At least that is what makes sense. If we modify the incoming request, it's not the incoming request anymore. And the outgoing request is a copy of the incoming request that was modified by the middlware, and with a ranked list of TES endpoints attached. And it is also the outgoing request that should be updated according to what is happening on the remote service and what is, essentially, returned to the user when they ask for info on that task (except a few properties like the start and other times pertaining to the task itself (not for the executor).

This all looks like it is actually done the other way around (or maybe I'm being stupid).

If that is the case, please open an issue to fix in another PR.

uniqueg (approved):
Thanks! There is one confusing issue. Please have a look and let me know what you think (or open an issue if you agree). Anyway, this doesn't really affect _this_ PR, so I will approve this.

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Actually, task_outgoing is the one that is untouched, and all the changes (middleware,  the final response etc) are done in task_incoming.

uniqueg: Yes, but it doesn't make sense. Don't you agree? If we modify the incoming request, it's not the incoming request anymore. Perhaps instead of naming them "incoming" and "outgoing", it might be better to call one of them just `task` and the other one `task_original`. Then `task` is the working copy which gets modified and always keeps all the info just as we want to return it to the user/client.

Ayush5120: If you want to rename, i.e task_incoming as task_outgoing and vice versa we can do that.

Ayush5120: > 

Yes makes sense, I am creating an issue for that where we just need to rename it.

Ayush5120: Created issue #137

Issue #136, opened 2023-02-12
Fix docstrings


Ayush5120:
- Follow Google-styled docstring standards.
- Check for accuracy and provide detailed docstring for functions/methods in the whole codebase.

Ayush5120:
Closing as completed in #138

Issue #137, opened 2023-02-16
Fix naming of TES task request objects


Ayush5120:
**Description** :

- The Tes task request, when forwarded to `proTES`, undergoes certain operations on the request body such as updating the task state, adding a task identifier, logging information, and specifying the TES endpoint where the task is forwarded, etc. All of these operations are currently performed on the `task_incoming` object, which is an instance of the `TesTask` class and where the final results of the executed task are stored.

- The `task_outgoing` object, on the other hand, is used to save and store the incoming request in the database without modifying it. This is done to keep a record of what the client had originally passed when the task was received.

**Problem** : 

- The naming of TES task requests, that is `task_incoming` and `task_outgoing` is currently confusing.

**Solution** :

- Rather than using the terms `task_incoming`and `task_outgoing`, it may be more appropriate to refer to one as simply `task` and the other as `task_original`. The `task` copy would be the working version that gets modified and retains all the necessary information to be returned to the user or client, while the `task_original` would be the one that remains unchanged.

BYZANTINE26:
Hi, 

I have gone through the Description and Solution of your issue, and understood the code base. 
I have identified 3 files where changes need to be made for task_incoming: `pro_tes\ga4gh\tes\models.py`, `pro_tes\ga4gh\tes\task_runs.py` and `pro_tes\tasks\track_task_progress.py`

For task_outgoing: `pro_tes\ga4gh\tes\models.py` and `pro_tes\ga4gh\tes\task_runs.py`

I request you to assign this issue to me.

Thanks

Ayush5120:
Thanks, @BYZANTINE26, Please read the contributor guidelines before and you are ready to go, I am assigning this to you.

PR #138, opened 2023-02-20
fix: docstring and minor refactoring

Ayush5120:
- Addressed #136.
- refactor to use list comprehension.
- Raised custom exception instead of using generic in the middleware module.

### DIFF
--- a/pro_tes/ga4gh/tes/models.py
+++ b/pro_tes/ga4gh/tes/models.py
@@ -666,8 +666,8 @@ class TesEndpoint(CustomBaseModel):
     Args:
         host: Host at which the TES API is served that is processing this
             request; note that this should include the path information but
-            *not* the base path path defined in the TES API specification;
-            e.g., specify https://my.tes.com/api if the actual API is hosted at
+            *not* the base path defined in the TES API specification; e.g.,
+            specify https://my.tes.com/api if the actual API is hosted a
             https://my.tes.com/api/ga4gh/tes/v1.
         base_path: Override the default path suffix defined in the TES API
             specification, i.e., `/ga4gh/tes/v1`.
@@ -676,8 +676,8 @@ class TesEndpoint(CustomBaseModel):
     Attributes:
         host: Host at which the TES API is served that is processing this
             request; note that this should include the path information but
-            *not* the base path path defined in the TES API specification;
-            e.g., specify https://my.tes.com/api if the actual API is hosted at
+            *not* the base path defined in the TES API specification; e.g.,
+            specify https://my.tes.com/api if the actual API is hosted at
             https://my.tes.com/api/ga4gh/tes/v1.
         base_path: Override the default path suffix defined in the TES API
             specification, i.e., `/ga4gh/tes/v1`.
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -387,7 +387,7 @@ class TaskRuns:
         """Sanitize request for use with py-tes.
 
         Args:
-            payloads: Request payload.
+            payload: Request payload.
 
         Returns:
             Sanitized request payload.
@@ -417,7 +417,7 @@ class TaskRuns:
         return payload
 
     def _set_projection(self, view: str) -> Dict:
-        """Set database projectoin for selected view.
+        """Set database projection for selected view.
 
         Args:
             view: View path parameter.
@@ -453,7 +453,17 @@ class TaskRuns:
     def _update_task(
         self, payload: dict, db_document: DbDocument, start_time: str, **kwargs
     ) -> DbDocument:
-        """Update the task object."""
+        """Update the task object.
+
+        Args:
+            payload: A dictionary containing the payload for the update.
+            db_document: The document in the database to be updated.
+            start_time: The starting time of the incoming TES request.
+            **kwargs: Additional keyword arguments passed along with request.
+
+        Returns:
+            DbDocument: The updated database document.
+        """
         logs = self._set_logs(
             payloads=deepcopy(payload), start_time=start_time
         )
@@ -467,7 +477,15 @@ class TaskRuns:
         return db_document
 
     def _set_logs(self, payloads: dict, start_time: str) -> Dict:
-        """Set up the logs for the request."""
+        """Create or update `TesTask.logs` and set start time.
+
+        Args:
+            payload: A dictionary containing the payload for the update.
+            start_time: The starting time of the incoming TES request.
+
+        Returns:
+            Task logs with start time set.
+        """
         if "logs" not in payloads.keys():
             logs = [
                 {
@@ -491,7 +509,16 @@ class TaskRuns:
         tes_uri: str,
         remote_task_id: str,
     ) -> DbDocument:
-        """Update the document in the database."""
+        """Set end time, task metadata in `TesTask.logs`, and update document.
+
+        Args:
+            db_connector: The database connector.
+            tes_uri: The TES URI where the task if forwarded.
+            remote_task_id: Task identifier at the remote TES instance.
+
+        Returns:
+            The updated database document.
+        """
         time_now = datetime.now().strftime("%m-%d-%Y %H:%M:%S")
         tes_endpoint_dict = {"host": tes_uri, "base_path": ""}
         db_document = db_connector.upsert_fields_in_root_object(
@@ -529,7 +556,18 @@ class TaskRuns:
     def _update_task_metadata(
         self, db_document: DbDocument, tes_uri: str, remote_task_id: str
     ) -> DbDocument:
-        """Update the task metadata."""
+        """Update the task metadata.
+
+        Set TES endpoint and remote task identifier in `TesTask.logs.metadata`.
+
+        Args:
+            db_document: The document in the database to be updated.
+            tes_uri: The TES URI where the task if forwarded.
+            remote_task_id: Task identifier at the remote TES instance.
+
+        Returns:
+            The updated database document.
+        """
         for logs in db_document.task.logs:
             tesNextTes_obj = TesNextTes(id=remote_task_id, url=tes_uri)
             if logs.metadata.forwarded_to is None:
--- a/pro_tes/middleware/middleware.py
+++ b/pro_tes/middleware/middleware.py
@@ -14,7 +14,16 @@ class AbstractMiddleware(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def modify_request(self, request):
-        """Modify the request before it is sent to the TES instance."""
+        """Modify the incoming task request.
+
+        Abstract method.
+
+        Args:
+            request: The incoming request object.
+
+        Returns:
+            The modified request object.
+        """
 
 
 class TaskDistributionMiddleware(AbstractMiddleware):
@@ -22,6 +31,7 @@ class TaskDistributionMiddleware(AbstractMiddleware):
 
     Attributes:
         tes_uri: TES instance best suited for TES task.
+        input_uris: A list of input URIs from the incoming request.
     """
 
     def __init__(self) -> None:
@@ -30,13 +40,19 @@ class TaskDistributionMiddleware(AbstractMiddleware):
         self.input_uris: List[str] = []
 
     def modify_request(self, request):
-        """Add ranked list of TES instances to request body.
+        """Modify the incoming task request.
+
+        Abstract method
 
         Args:
             request: Incoming request object.
 
         Returns:
-            Tuple of modified request object.
+            The modified request object.
+
+        Raises:
+            pro_tes.exceptions.NoTesInstancesAvailable: If no valid TES
+                instances are available.
         """
         if "inputs" in request.json.keys():
             for index in range(len(request.json["inputs"])):
--- a/pro_tes/middleware/middleware.py
+++ b/pro_tes/middleware/middleware.py
@@ -3,6 +3,7 @@
 import abc
 from typing import List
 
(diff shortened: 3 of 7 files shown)


pro_tes/ga4gh/tes/models.py
```diff
@@ -666,7 +666,7 @@ class TesEndpoint(CustomBaseModel):
     Args:
         host: Host at which the TES API is served that is processing this
             request; note that this should include the path information but
-            *not* the base path path defined in the TES API specification;
+            *not* the base path defined in the TES API specification;
```
uniqueg: Make use of all 80 chars in each line of a paragraph. After deleting the repeat `path`, you can move the `e.g.,` to the previous line.

Btw, there are VS Code plugins that can do this automatically (and that you can configure with the desired line length).

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -453,7 +453,16 @@ def _set_projection(self, view: str) -> Dict:
     def _update_task_incoming(
         self, payload: dict, db_document: DbDocument, start_time: str, **kwargs
     ) -> DbDocument:
-        """Update the task incoming object."""
+        """Update the task incoming object.
```
uniqueg: Better: "the incoming task document".

However, as the incoming task document is supposed to be immutable, once created (can you check if there is a MongoDB setting that we could use here to enforce this?), you will anyway need to change that in an upcoming PR. Still, better to correct this now so that it doesn't propagate to the next PR.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -453,7 +453,16 @@ def _set_projection(self, view: str) -> Dict:
...
+
+        Args:
+            payload: A dictionary containing the payload for the update.
+            db_document: The document in the database to be updated.
+            start_time: The starting time of the incoming tes request.
```
uniqueg: Please use `TES` (capitalized) instead of `tes`. Please update this in _all_ docstrings.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -467,7 +476,15 @@ def _update_task_incoming(
         return db_document
 
     def _set_logs(self, payloads: dict, start_time: str) -> Dict:
-        """Set up the logs for the incoming request."""
+        """Set up the logs for the incoming request.
```
uniqueg: Better (more precies): "Create or update `TesTask.logs` and set start time"

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -467,7 +476,15 @@ def _update_task_incoming(
...
+
+        Args:
+            payloads: A dictionary containing the payload for the update.
```
uniqueg: It should be `payload` (singular) not `payloads`.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -467,7 +476,15 @@ def _update_task_incoming(
...
+            start_time: The starting time of the incoming tes request.
+
+        Returns:
+            Dict: The updated logs.
```
uniqueg: Better (more precise): Task logs with start time set.

Since we have type hints in function/method signatures, please do not repeat them in docstrings. So please drop `Dict:`. Check other instances, too, including `Args:` and `Attributes:` sections.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -491,7 +508,16 @@ def _update_doc_in_db(
         tes_uri: str,
         remote_task_id: str,
     ) -> DbDocument:
-        """Update the document in the database."""
+        """Update the document in the database.
```
uniqueg: Better (more concise): "Update the database document". Also, briefly looking at the code, this does not look like a _generic_ database document update method. Rather, you are updating something specific here, namely information about the TES endpoint that the task is forwarded to (going to be forwarded to or _has been_ forwarded to?), and a couple of other things.

Please try to be more descriptive/precise. What kind of update does this method really do?

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -491,7 +508,16 @@ def _update_doc_in_db(
...
+
+        Args:
+            db_connector: The database connector.
+            tes_uri: The tes uri where the task if forwarded.
```
uniqueg: Please capitalize both "TES" and "URI" (here and in all other docstrings).

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -529,7 +555,16 @@ def _update_doc_in_db(
     def _update_task_metadata(
         self, db_document: DbDocument, tes_uri: str, remote_task_id: str
     ) -> DbDocument:
-        """Update the task metadata."""
+        """Update the task metadata.
```
uniqueg, line 559: Again, more details would be nice. How does this differ from the previous function? Isn't all the data metadata, really?

Try to find a succinct (not more than one line!) way of summarizing the function. Then, if you feel that you need to describe more, you can add an extended description (leave one line blank after the summary, and before the "Args:" section).

pro_tes/middleware/middleware.py
```diff
@@ -13,14 +14,22 @@ class AbstractMiddleware(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def modify_request(self, request):
-        """Modify the request before it is sent to the TES instance."""
+        """Abstract method to modify the incoming request.
```
uniqueg: Docstring summaries should start with a verb, in imperative form. So better to use:

```
"""Modify the incoming task request.

Abstract method.

Args:
    ...
"""
```

pro_tes/middleware/middleware.py
```diff
@@ -13,14 +14,22 @@ class AbstractMiddleware(metaclass=abc.ABCMeta):
...
+            The modified request object.
+        """
 
 
 class TaskDistributionMiddleware(AbstractMiddleware):
     """Inject task distribution logic.
 
     Attributes:
         tes_uri: TES instance best suited for TES task.
+        input_uris: A list of input URIs in the incoming request.
```
uniqueg: Better "[...] _from_ the incoming request"

pro_tes/middleware/middleware.py
```diff
@@ -35,7 +44,7 @@ def modify_request(self, request):
             request: Incoming request object.
 
         Returns:
-            Tuple of modified request object.
+            The modified request object.
         """
```
uniqueg, line 56: Make sure that in all methods/functions that _explicitly_ raise an error, there is a `Raises:` section that describes what error is raised and under what circumstances. Here, for example:

```
"""...

Args:
    ...

Returns:
    ...

Raises:
    `pro_tes.exceptions.NoTesInstancesAvailable`: If no valid TES instances are
        available.
"""
```

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -71,7 +69,21 @@ def get_uri_combination(
     Args:
         input_uri: List of input uris of TES request.
         tes_uri: List of TES instance.
+
     Returns:
+        An AccessUriCombination object, consisting of the following keys:
```
uniqueg: Refer to `AccessUriCombination` via a fully qualified path, e.g., `pro_tes.models.AccessUriCombination` (I didn't check this, there may be other submodules). Then there is no need to described that object, people can just look it up.

Even better: you can use Sphinx shortcuts to refer to other functions, methods, modules etc: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

This is really powerful!

In this case, it would be something like:

```markdown
:class:`pro_tes.models.AccessUriCombination` 
```

(you can omit the `:py` at the start, see https://stackoverflow.com/a/22714510)

including backticks and all.

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -153,17 +171,21 @@ def ip_distance(
 ) -> Dict[str, Dict]:
     """Compute ip distance between ip pairs.
```
uniqueg: Please capitalize "IP"

uniqueg (changes requested):
Thanks @Ayush5120!

Please split up into three PRs here, in this order:
- feat: return specific error if no TES available
- refactor: minor improvements on code clarity
- docs: extend and correct docstrings

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: This function sets the end time( in TesTask.logs), sets the TES endpoint where the task is forwarded to and its remote task id ( in TesTask.logs.metadata), and then finally inserts the updated document in DB.

Ayush5120: will update its docstring accordingly in next commit.

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: At this stage, the incoming task document will keep updating till the task is completed at the remote TES, by the way, will fix this naming issue in #137. so will it make sense to make it immutable? correct me if I am wrong

uniqueg: Fine to address it in #137, but _then_ the incoming task document should indeed be immutable (i.e., the parts that are provided by the client, not the . It's a log/reference to what the user sent, we need this for accounting/audit purposes. We then create the _outgoing_ request and _that_ we can modify according to whatever middleware is configured to act upon the incoming request.

Ayush5120: Okay, understood but I don't think it is good to fix this in this pr as the naming of the task document is already messed up, we can do this once the naming issue #137 is resolved, if you agree I will create an issue for this.

pro_tes/ga4gh/tes/models.py
Ayush5120: Done

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Issue Created #142

pro_tes/middleware/task_distribution/distance.py

pro_tes/ga4gh/tes/task_runs.py

pro_tes/ga4gh/tes/task_runs.py

pro_tes/ga4gh/tes/task_runs.py

pro_tes/ga4gh/tes/task_runs.py

pro_tes/middleware/middleware.py
Ayush5120: Done

pro_tes/middleware/task_distribution/distance.py

pro_tes/ga4gh/tes/task_runs.py

pro_tes/middleware/middleware.py
Ayush5120: Done

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Done

pro_tes/middleware/middleware.py
uniqueg: Not really. It doesn't start with a verb in imperative form. I mean it does, technically, because "abstract" can also be a verb, but here it is meant as an adjective. I have already written how to write it instead. Just use that.

pro_tes/ga4gh/tes/task_runs.py
uniqueg: Yup. But please add a `.` at the end of the description. Also, it would be good to put backticks around any expressions like `TesTask.logs.metadta`.

pro_tes/ga4gh/tes/task_runs.py

pro_tes/middleware/middleware.py
Ayush5120: Done

pro_tes/middleware/middleware.py
Ayush5120: ooh Sorry, for that now fixed

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Fixed

pro_tes/middleware/middleware.py
```diff
@@ -29,7 +29,7 @@ class TaskDistributionMiddleware(AbstractMiddleware):
 
     Attributes:
         tes_uri: TES instance best suited for TES task.
-        input_uris: A list of input URIs in the incoming request.
+        input_uris: [...] from the incoming request.
```
uniqueg: Sorry, I meant "A list of input URIs from the incoming request" :)

`[...]` I use as a placeholder for everything that comes before or after the expression I want to quote (so that I don't have to repeat the whole thing) ;-)

pro_tes/ga4gh/tes/task_runs.py
uniqueg: And the backticks? Put:

```console
[...] expressions like `TesTask.logs.metadata`.
```

pro_tes/middleware/middleware.py
Ayush5120: Oh, I should have thought that, will fix in the next commits

Ayush5120: Fixed

pro_tes/ga4gh/tes/task_runs.py
Ayush5120: Done

uniqueg (approved):

PR #141, opened 2023-03-01
refactor: naming of TES tasks DB documents

### LINKED ISSUE #137: Fix naming of TES task request objects

BYZANTINE26:
Resolves #137

### DIFF
--- a/pro_tes/ga4gh/tes/models.py
+++ b/pro_tes/ga4gh/tes/models.py
@@ -692,24 +692,24 @@ class DbDocument(CustomBaseModel):
     """Create model instance for task request database document.
 
     Args:
-        task_incoming: Information about incoming task.
-        task_outgoing: Information about outgoing task.
+        task: Information about task.
+        task_original: Information about original task.
         user_id: Identifier of resource owner.
         worker_id: Identifier of worker task.
         basic_auth: Basic authentication credentials.
         tes_endpoint: External TES endpoint.
 
     Attributes:
-        task_incoming: Information about incoming task.
-        task_outgoing: Information about outgoing task.
+        task: Information about task.
+        task_original: Information about original task.
         user_id: Identifier of resource owner.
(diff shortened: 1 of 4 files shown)


Ayush5120:
Thanks, @BYZANTINE26, for raising the PR. I have seen the changes, and they appear to be fine to me. However, it would be helpful if @uniqueg  could add comments to it.

uniqueg (approved):
Thanks a lot @BYZANTINE26, this looks solid.

uniqueg:
Thanks @BYZANTINE26. Note that I have renamed the PR (changed `fix` to `refactor`, as nothing was actually _fixed_ - the outside behavior of the app didn't change; made it a bit more descriptive) and also edited your PR description so that GitHub actually picks up the link to the issue and will automatically close it upon merging (you can't have any words between "resolves" and the issue number). Check here for details: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #143, opened 2023-03-08
fix: premature system error state

### LINKED ISSUE #139: Task state set incorrectly when best TES instance fails
**Description** :

- `proTES` tries to send the task to the best available TES which is done using a ranked list of TES instances, but when the submission of the task to the best available TES fails, it tries to forward the task to the next available instances.

**Problem**:

- When submission of the task to the best TES instance fails the state is set to `SYSTEM_ERROR`.Ideally, it should not set the state to `SYSTEM_ERROR` until the task submission fails on all the available TES instances.

SohamRatnaparkhi:
**IMPORTANT: Please create an issue before filing a pull request! Changes need to be discussed before proceeding. Please read the [contribution guidelines](CONTRIBUTING.md).**

**Details**

Resolves #139 

When submission of the task to the best TES instance fails the state is set to SYSTEM_ERROR.Ideally, it should not set the state to SYSTEM_ERROR until the task submission fails on all the available TES instances.

To solve this, I _moved the code-block_ which was specifically related to setting the error state to `SYSTEM ERROR` to the place  which will only be touched if all the TES instances fail. But for every failed instance, a message will be logged.

**Testing**

NA

**Documentation**

NA

**Style**

My code adheres to coding convention of this project.

**Closing issues**

closes issue: #139

### DIFF
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -209,9 +209,12 @@ class TaskRuns:
             )
             return {"id": db_document.task.id}
 
-        # No suitable TES instance found. Task state set to 'SYSTEM_ERROR'.
         db_connector.update_task_state(
             state=TesState.SYSTEM_ERROR.value
+        )
+        logger.error(
+            "No suitable TES instance found. Task state set to "
+            "'SYSTEM_ERROR'."
         )
 
     def list_tasks(self, **kwargs) -> Dict:
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -209,7 +209,7 @@ class TaskRuns:
             )
             return {"id": db_document.task_incoming.id}
 
-        # set TES state to SYSTEM_ERROR as task submission failed on all the available TES instances.
+        # No suitable TES instance found. Task state set to 'SYSTEM_ERROR'.
         db_connector.update_task_state(
             state=TesState.SYSTEM_ERROR.value
         )
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -127,9 +127,7 @@ class TaskRuns:
                     password=db_document.basic_auth.password,
                 )
             except ValueError as exc:
-                db_connector.update_task_state(
-                    state=TesState.SYSTEM_ERROR.value
-                )
+
                 logger.info(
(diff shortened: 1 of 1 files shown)


pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -213,6 +209,11 @@ def create_task(  # pylint: disable=too-many-statements,too-many-branches
             )
             return {"id": db_document.task.id}
 
+        # set TES state to SYSTEM_ERROR as task submission failed on all the available TES instances.
```
uniqueg: Would be good to add another log message here indicating something like: `No suitable TES instance found. Task state set to 'SYSTEM_ERROR'.`

uniqueg (changes requested):
Nice, thanks a lot. One minor suggestion.

pro_tes/ga4gh/tes/task_runs.py
uniqueg: This line is too long, lines should be <80 chars.

SohamRatnaparkhi: Okay. I have made the required changes in the next commit

SohamRatnaparkhi: Okay. I have replaced my comment with `No suitable TES instance found. Task state set to 'SYSTEM_ERROR'.` in the next commit.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -209,7 +209,7 @@ def create_task(  # pylint: disable=too-many-statements,too-many-branches
             )
             return {"id": db_document.task.id}
 
-        # set TES state to SYSTEM_ERROR as task submission failed on all the available TES instances.
+        # No suitable TES instance found. Task state set to 'SYSTEM_ERROR'.
```
uniqueg: I'm sorry if my comment was not clear. What I meant to say is that it would be good to have a logging message to that effect, something like:

```python
            logger.error(
                "No suitable TES instance found. Task state set to "
                "'SYSTEM_ERROR'."
            )
```

You can then also remove the comment, because it will be obvious.

uniqueg (changes requested):

pro_tes/ga4gh/tes/task_runs.py
SohamRatnaparkhi: Okay, got it! Will update these logs in next commit.

uniqueg (approved):

PR #144, opened 2023-03-08
feat: raise 400/500 for invalid input/TES URIs

### LINKED ISSUE #134: Several issues with TES/input URI processing in distance-based task distribution logic

Ayush5120:
Partly addresses #134


uniqueg (approved):

uniqueg:
Good stuff!

PR #145, opened 2023-03-10
feat: filter tasks by name prefix

### LINKED ISSUE #127: Tasks not filtered by name prefix
A `GET` request to `/tasks?name_prefix=foo&view=MINIMAL` should return only tasks whose identifiers start with `foo`. However, it appears that all tasks are returned, not filtered at all.

**To reproduce**

* `POST` one or more tasks to a given proTES instance
* Store URL to proTES TES API in variable `TES` (e.g., `https://my.tes.org/ga4gh/tes/v1`)
* Run the following command
   ```bash
   curl -X GET "${TES}/tasks?name_prefix=foo&page_size=10&view=MINIMAL" -H "accept: application/json"
   ```

**Expected behavior**

Should only list tasks whose IDs start with `foo` (probably none at all).

**Actual behavior**

Lists all tasks, including those whose identifiers do _not_ start with `foo`.

SohamRatnaparkhi:
**IMPORTANT: Please create an issue before filing a pull request! Changes need to be discussed before proceeding. Please read the [contribution guidelines](CONTRIBUTING.md).**

**Details**

Solves issue #127

Currently, when name_prefix is passed as a parameter with `\task` route, the tasks aren't filtered. This PR involves adding a wildcard regular expression matching and filtering of tasks with respect to name_prefix (if present).
**Note** : 
Filtering is strongly **case sensitive**. 
Therefore, if name_prefix="Foo" then tasks with their names starting with "Foo such as Foo1, Foo2 (Foo*) will be considered but (foo, FOo, etc.) wont be considered.

**Testing**
NA

**Documentation**
NA

**Style**

Code adheres to styling of the project.

**Closing issues**

Closes #127 

**Credit**

Add your credentials to the [list of contributors](contributors.md) once your pull request was merged.

### DIFF
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -240,6 +240,13 @@ class TaskRuns:
         view = kwargs.get("view", "BASIC")
         projection = self._set_projection(view=view)
 
+        name_prefix = kwargs.get("name_prefix")
+
+        if name_prefix is not None:
+            filter_dict["task_original.name"] = {
+                "$regex": f"^{name_prefix}"
+            }
+
         cursor = (
             self.db_client.find(filter=filter_dict, projection=projection)
             .sort("_id", -1)
--- a/pro_tes/ga4gh/tes/task_runs.py
+++ b/pro_tes/ga4gh/tes/task_runs.py
@@ -240,7 +240,7 @@ class TaskRuns:
         view = kwargs.get("view", "BASIC")
         projection = self._set_projection(view=view)
 
-        name_prefix = kwargs.get("name_prefix")
+        name_prefix: str = kwargs.get("name_prefix")
 
         if name_prefix is not None:
             filter_dict["task_original.name"] = {


uniqueg:
Thanks @SohamRatnaparkhi :)

The [TES specification does not say anything about case sensitivity](https://github.com/ga4gh/task-execution-schemas/blob/1df37d34242f74d3f03475c6b9de3324b8094054/openapi/task_execution_service.openapi.yaml#L99-L105), which is a pity. However, without any other information, I would interpret that to mean that the filtering _should_ be case sensitive.

Request a review when you are ready.

SohamRatnaparkhi:
> Thanks @SohamRatnaparkhi :)
> 
> The [TES specification does not say anything about case sensitivity](https://github.com/ga4gh/task-execution-schemas/blob/1df37d34242f74d3f03475c6b9de3324b8094054/openapi/task_execution_service.openapi.yaml#L99-L105), which is a pity. However, without any other information, I would interpret that to mean that the filtering _should_ be case sensitive.
> 
> Request a review when you are ready.

Okay cool! Then I am requesting a review with case sensitive regex.
> In future if we need to make it case insensitive, then we just need to add `/i` or a `{ option : 'i' }` to regex/filter_dict

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -240,6 +240,13 @@ def list_tasks(self, **kwargs) -> Dict:
         view = kwargs.get("view", "BASIC")
         projection = self._set_projection(view=view)
 
+        name_prefix = kwargs.get("name_prefix")
```
uniqueg: For clarity, you could add a type hint for `name_prefix`, like so:

```python
name_prefix: str = ...
```

uniqueg (changes requested):
Looks great, very clean :) Just one minor suggestion.

@Ayush5120: Once @SohamRatnaparkhi has added the type hint, can you please test the feature and, if it works, approve the PR? Thanks!

pro_tes/ga4gh/tes/task_runs.py
SohamRatnaparkhi: Okay, will add it in my next commit

Ayush5120: > Looks great, very clean :) Just one minor suggestion.
> 
> @Ayush5120: Once @SohamRatnaparkhi has added the type hint, can you please test the feature and, if it works, approve the PR? Thanks!
Sure, I will do that

Ayush5120:
Thanks, @SohamRatnaparkhi I have tested it and it works fine.

uniqueg (approved):

PR #147, opened 2023-03-13
fix: exception handling in IP distance calculation

### LINKED ISSUE #146: Error handling issues in calculate_distance function of distance-based task distribution module
Extension of issue #134 
The `calculate_distance` function in the distance-based task distribution module has an issue with exception handling, leading to unexpected behavior or errors. Specifically:

- `ip_distance` function is called to calculate distances between all IPs. However, if it raises a `ValueError`, the exception is ignored, which could lead to incorrect results or unexpected behavior if the distances_full variable is used later in the code.
- Similarly, a `KeyError` exception is ignored when trying to access the distances_full dictionary, which might lead to incorrect outputs.

To solve this issue handle these exceptions more robustly, for example, by logging the exception or raising an error to alert the user that something has gone wrong.

Ayush5120:
- Resolves #146

### DIFF
--- a/pro_tes/middleware/task_distribution/distance.py
+++ b/pro_tes/middleware/task_distribution/distance.py
@@ -249,7 +249,7 @@ def calculate_distance(
                 ]
             except KeyError as exc:
                 raise KeyError(
-                     f"Distance not found for IP addresses: {ip_tuple}"
+                     f"Distances not found for IP addresses: {ip_tuple}"
                 ) from exc
 
     # Reshape distances keys for logging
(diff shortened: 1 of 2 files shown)


pro_tes/middleware/task_distribution/distance.py
```diff
@@ -242,8 +247,10 @@ def calculate_distance(
                 distances_unique[ip_tuple] = distances_full["distances"][
                     ip_tuple
                 ]
-            except KeyError:
-                pass
+            except KeyError as exc:
+                raise KeyError(
+                     f"Distance not found for IP addresses: {ip_tuple}"
+                ) from exc
```
uniqueg: "Distances not found for IP addresses" (distances in plural)

uniqueg (changes requested):
Just a minor typo, otherwise LGTM

pro_tes/middleware/task_distribution/distance.py

uniqueg (approved):

PR #148, opened 2023-03-15
feat: fall back to random distribution

### LINKED ISSUE #134: Several issues with TES/input URI processing in distance-based task distribution logic

Ayush5120:
Fall back to random task distribution if distance-based task distribution fails.

Partially fixes #134

### DIFF
--- a/pro_tes/middleware/middleware.py
+++ b/pro_tes/middleware/middleware.py
@@ -66,17 +66,15 @@ class TaskDistributionMiddleware(AbstractMiddleware):
                         request.json["inputs"][index]["url"]
                     )
 
-        if self.input_uris:
-            try:
-                self.tes_uris = distance.task_distribution(self.input_uris)
-            except (
-                TesUriError,
-                InputUriError,
-                IPDistanceCalculationError,
-                KeyError,
-            ):
-                self.tes_uris = random.task_distribution()
-        else:
+        try:
+            self.tes_uris = distance.task_distribution(self.input_uris)
+        except (
+            TesUriError,
+            InputUriError,
+            IPDistanceCalculationError,
+            KeyError,
+            ValueError
+        ):
             self.tes_uris = random.task_distribution()
 
         if self.tes_uris:
--- a/pro_tes/middleware/middleware.py
+++ b/pro_tes/middleware/middleware.py
@@ -3,7 +3,12 @@
 import abc
 from typing import List
 
-from pro_tes.exceptions import NoTesInstancesAvailable
+from pro_tes.exceptions import (
+    NoTesInstancesAvailable,
+    TesUriError,
+    InputUriError,
+    IPDistanceCalculationError,
+)
 from pro_tes.middleware.task_distribution import distance, random
 
 # pragma pylint: disable=too-few-public-methods
@@ -62,7 +67,15 @@ class TaskDistributionMiddleware(AbstractMiddleware):
                     )
 
         if self.input_uris:
-            self.tes_uris = distance.task_distribution(self.input_uris)
+            try:
+                self.tes_uris = distance.task_distribution(self.input_uris)
+            except (
+                TesUriError,
+                InputUriError,
+                IPDistanceCalculationError,
+                KeyError,
+            ):
+                self.tes_uris = random.task_distribution()
         else:
             self.tes_uris = random.task_distribution()
 
--- a/pro_tes/middleware/task_distribution/distance.py
+++ b/pro_tes/middleware/task_distribution/distance.py
@@ -37,18 +37,19 @@ def task_distribution(input_uri: List) -> List:
     Returns:
         A list of ranked TES instances, ordered by the minimum distance
         between the input files and each TES instance.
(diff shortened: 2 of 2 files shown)


pro_tes/middleware/task_distribution/distance.py
```diff
@@ -44,15 +44,22 @@ def task_distribution(input_uri: List) -> List:
     access_uri_combination = get_uri_combination(input_uri, tes_uri)
 
     # get the combination of the tes ip and input ip
-    ips = ip_combination(input_uri=input_uri, tes_uri=tes_uri)
+    try:
+        ips = ip_combination(input_uri=input_uri, tes_uri=tes_uri)
+    except (InputUriError, TesUriError):
+        return random.task_distribution()
+
```
uniqueg: I think the different middlewares should not be coupled. So I think this should be implemented _outside_ of `distance.py`, i.e., raise the error here, then catch it in the calling context and use the alternative middleware instead.

uniqueg (changes requested):

pro_tes/middleware/middleware.py
```diff
@@ -62,7 +67,15 @@ def modify_request(self, request):
                     )
 
         if self.input_uris:
```
uniqueg: What happens in `distance.task_distribution()` if `self.input_uris` is empty? Will it raise one of the errors you are catching?

If so (and it probably should!), there's no point in doing the `if`-`else` anymore - you can just do the `try` and `except`.

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -46,8 +45,10 @@ def task_distribution(input_uri: List) -> List:
     # get the combination of the tes ip and input ip
     try:
         ips = ip_combination(input_uri=input_uri, tes_uri=tes_uri)
-    except (InputUriError, TesUriError):
-        return random.task_distribution()
+    except InputUriError as exc:
+        raise InputUriError from exc
+    except TesUriError as exc:
+        raise TesUriError from exc
```
uniqueg: There's no point in doing that. Just let the error raise itself, implicitly! That is, remove the entire `try`-`except` here. Then you also do not need to document the exceptions in `Raises` in the docstring (which otherwise you would have needed to add, try to remember that).

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -58,8 +59,10 @@ def task_distribution(input_uri: List) -> List:
     # Calculate distances between all IPs
     try:
         distances = calculate_distance(ips_unique, tes_uri)
-    except (IPDistanceCalculationError, KeyError):
-        return random.task_distribution()
+    except IPDistanceCalculationError as exc:
+        raise IPDistanceCalculationError from exc
+    except KeyError as exc:
+        raise KeyError from exc
```
uniqueg: Same as above, just remove the `try`-`except`.

uniqueg (changes requested):
We're getting there :)

pro_tes/middleware/task_distribution/distance.py
Ayush5120: Understood, will fix it in the next commit.

pro_tes/middleware/middleware.py
Ayush5120: No the errors mentioned, do not respond to empty input URI, there was just an outside check, I can remove the outside check and maybe raise a `ValueError` in `distance.py` with a message ( "No input URI available) and then catch the error hence, removing the if-else block.

uniqueg: Yup. It doesn't make sense to use the distance-based logic if there are no inputs, so I think it would be good to raise a `ValueError` and then catch that outside and remove the `if`-`else`.

pro_tes/middleware/task_distribution/distance.py

pro_tes/middleware/task_distribution/distance.py

pro_tes/middleware/middleware.py
Ayush5120: Okay, fixed in the last commit.

pro_tes/middleware/task_distribution/distance.py
```diff
@@ -37,18 +37,19 @@ def task_distribution(input_uri: List) -> List:
...
+    Raises:
+        ValueError: If no input URIs are available.
     """
     foca_conf = current_app.config.foca
     tes_uri: List[str] = foca_conf.tes["service_list"]
     access_uri_combination = get_uri_combination(input_uri, tes_uri)
 
+    if not input_uri:
+        raise ValueError("No input URIs available.")
+
```
uniqueg: For performance reasons, this should come before anything else in this method. Especially because you are already using `input_uri` in the previous statement.

uniqueg (changes requested):
We're getting there. Just one very small change left :)

pro_tes/middleware/task_distribution/distance.py

uniqueg (approved):

PR #150, opened 2023-03-18
test: unit tests distribution middleware, distance

### LINKED ISSUE #15: Add unit tests
**Is your feature request related to a problem? Please describe.**

Add extensive unit tests for the entire app.

**Describe the solution you'd like**

Write unit tests (preferentially using [`pytest`](https://docs.pytest.org/en/7.2.x/)) for every class/method/function, aiming for 100% code coverage.

**Describe alternatives you've considered**
N/A

### LINKED ISSUE #134: Several issues with TES/input URI processing in distance-based task distribution logic

Ayush5120:
Unit test for the distance-based task distribution module with 100% coverage.

Resolves #134 and Partially resolves #15 

```
Name                                               Stmts   Miss  Cover
----------------------------------------------------------------------
pro_tes/middleware/task_distribution/distance.py     100      0   100%
----------------------------------------------------------------------
TOTAL                                                100      0   100%
```

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -49,6 +49,9 @@ jobs:
       - name: Run integration tests
         shell: bash
         run: pytest tests/test_integration
+      - name: Run unit tests
+        shell: bash
+        run: pytest tests/unitTest/pro_tes/middleware
       - name: Tear down app
         run: docker-compose down
   publish:
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 foca>=0.12.0
 gunicorn>=20.1.0,<21
 py-tes>=0.4.2
+pytest-ordering>=0.6
 ip2geotools>=0.1.6
 geopy>=2.2.0
 types-PyYAML>=6.0.11
--- a/tests/unitTest/pro_tes/middleware/mock_data.py
+++ b/tests/unitTest/pro_tes/middleware/mock_data.py
@@ -1,0 +1,359 @@
+"""Mock data for Testing."""
+
+from pydantic import AnyUrl, HttpUrl
+
+from pro_tes.middleware.models import (
+    AccessUriCombination,
+    TaskParams,
+    TesDeployment,
+    TesStats
+)
+
+INDEX_CONFIG_TASKS = {"keys": [("task_id", 1), ("worker_id", 1)]}
+
+COLLECTION_CONFIG_TASKS = {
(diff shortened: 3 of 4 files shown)


.github/workflows/checks.yaml
```diff
@@ -49,6 +49,9 @@ jobs:
       - name: Run integration tests
         shell: bash
         run: pytest tests/test_integration
+      - name: Run unit tests
+        shell: bash
+        run: pytest tests/unitTest/pro_tes/middleware
```
uniqueg, line 54: If you name your directories and files `*_test` or `test_*`, `pytest` will autodiscover all tests and you don't need to specify them on the command line. See here for details: https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html

In any case, Python naming conventions dictate snake case, so we should stick to that.

.github/workflows/checks.yaml
```diff
@@ -49,6 +49,9 @@ jobs:
...
       - name: Tear down app
         run: docker-compose down
```
uniqueg, line 56: Is this really not indented? Looks like that to me on my review screen on the phone...

uniqueg (changes requested):
Please change the paths for the test files according to Python and `pytest` conventions. Have a look at FOCA for examples. Otherwise looks good :)

.github/workflows/checks.yaml
Ayush5120: This is well Indented.

.github/workflows/checks.yaml
Ayush5120: Actually, there are some tests that are pretty old and need to be rewritten (which are at `tests/unitTest/pro_tes/endpoints`) otherwise they will always fail, that's why excluded them by mentioning only the required tests to run ( i.e. inside `tests/unitTest/pro_tes/middleware`).
What I can do now is rename the directories as you mentioned, now, but once all the test inside the unitTest directory is running fine, we can then change the command line to  `pytest tests/unittest/`

uniqueg: Oh, I didn't know we _had_ unit tests :) Please fix them ASAP and rename the packages/modules according to Python/`pytests` naming conventions and update the `pytest` call as you say :pray:

uniqueg (approved):
Oh, I didn't know we already _had_ unit tests :)

Please fix these ASAP, rename the packages/modules according to Python/`pytest` naming conventions and update the pytest call as you say :pray:

PR #160, opened 2023-06-19
feat: sequential execution of middleware

### LINKED ISSUE #124: Config param to choose task distribution logic
Check https://github.com/elixir-cloud-aai/proTES/pull/119#discussion_r1083449941 for context.

A good solution would keep the middleware handler truly generic. Each middleware should follow one of a number of abstract signatures. At least the following ones should be covered:

1. Change the request itself (`tesTask` in, `tesTask` out)
2. Change the request destination, i.e., task distribution logic: `tesTask` and list of known TES instances in, ranked list of TES instances out

The user would then list a number of middleware functions (either part of proTES or external ones) in the config, for these two (and possibly other, if we can think of any) signatures/sections. The handler would then apply these in the listed order. For example:

```yaml
middleware:
    mutate_request:  # signature 1. above
        - some_external_middlware.harmonized_tes_requests
        - some_external_middlware.tes_v1_0_0_to_v1_1_0
    mutate_destination:  # signature 2. above
        - some_external_middleware.discard_unauthorized_tes_instances
        - pro_tes.middleware.task_distribution.distance
```

Here, the handler would mutate the request by first harmonizing incoming requests, then converting any v1.0.0 requests to v1.1.0 (obviously, these are just examples of what middleware could hypothetically do). The handler would also set the destination by first discarding any TES instances that the user anyway doesn't have access to, then use the built-in distance task distribution logic to find the available/accessible TES that is closest to the input data.

Ideally, the handler would also check that the signature of the provided functions fit.

Other than that, there should be no specific logic in the handler (which there currently is), and no mention of any specific, hard-coded middleware (which there also is).

It might be useful to check if/how abstract classes/interfaces are done in Python to specify and validate the function signatures.

Ayush5120:
* Support for sequential execution of middleware to be injected into incoming `POST /tasks` requests
* Middleware classes to be executed can be specified, in fully qualified docstring notation, in the app configuration, in the desired order of execution, e.g.:
  ```yaml
  middlewares:
    - "package.subpackage.module.MiddlewareClass1"
    - "package.subpackage.module.MiddlewareClass2"
* It is possible to specify one or more "fallback middlewares", in case the desired middleware fails; to achieve that, list nesting by one level is supported, such that each inner list is treated as a single "middleware group", from which at most one middleware is applied - the first one that does not raise a `MiddlewareException`; in the following example, proTES first attempts to apply `MiddlewareClass1a`; if it works, it will advance to the next item in the outer list, `MiddlewareClass2` and try to apply that; however, if `MiddlewareClass1a` raises an error, proTES will first try to apply `MiddlewareClass1b` before progressing to `MiddlewareClass2`:
  ```yaml
  middlewares:
    - - "package.subpackage.module.MiddlewareClass1a"
      - "package.subpackage.module.MiddlewareClass1b"
    - "package.subpackage.module.MiddlewareClass2"
  ```
* If all middlewares of any middleware group fail (including in groups of size 1, i.e, without list nesting), proTES will respond with a `500` response
* Developers can implement their own middlewares by implementing the `pro_tes.middleware.abstract_middleware.AbstractMiddlewareClass`; of course, any middleware class provided via external packages needs to be made available to the application (e.g., by adding the corresponding package to `requirements.txt` and rebuilding the app and worker images)
* Available middlewares for task distribution (random and distance-based) are now available in `pro_tes.plugins.middlewares.task_distribution` with all their dependencies
* Task distribution middlewares are applied by default, as a single middleware group, with the distance-based distribution (`pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance`) being attempted first, and with the random-based task-distribution (`pro_tes.plugins.middlewares.task_distribution.random.TaskDistributionRandom`) serving as a fallback; this behavior can be changed by modifying the `middlewares` app configuration section
* Developers can optionally make use of the base class `pro_tes.plugins.middlewares.task_distribution.base.TaskDistributionBaseClass` for implementing other task distribution middlewares

Minor changes:
* Replace legacy (Python <=3.7) Type hints for `List`s, `Dict`s and `Tuple`s with their lowercase counterparts across the entire application
* Silence `mypy` complaints about missing type hints/stubs in external packages
* Drastically simplify middleware handler and task distribution middleware code
* Streamline `POST /tasks` controller code as a preparatory step for a major refactoring
* Streamline logging
* Temporarily disable unit tests in preparation for an effort to write consistent unit tests for the entire application

Closes #124


pro_tes/config.yaml
```diff
@@ -140,3 +140,8 @@ tes:
 
 storeLogs:
   execution_trace: True
+
+middlewares:
+  mutate_destination:
+    - - 'pro_tes.middleware.middleware.DistanceTaskDistribution'
+      - - 'pro_tes.middleware.middleware.RandomTaskDistribution'
```
uniqueg: Hmmm, I'm actually not too happy with this structure (and syntax).
I think having a list structure like this would make more sense:

```py
[[mw1, mw1_fb1, mw1_fb2], [mw2, mw2_fb1], mw3]
```

So have one outer list item for each middleware group, where each item can either be a list of options tried in sequential order until the first one passes _or_ a plain string (if there are no fallbacks).

The above in YAML would then become:

```yaml
- - mw1
  - mw1_fb1
  - mw1_fb2
- - mw2
  - mw2_fb1
- mw3
```

OR

```yaml
- [mw1, mw1_fb1, mw1_fb2]
- [mw2, mw2_fb1]
- mw3
```

For this example it would be:

```yaml
middlewares:
   - - 'pro_tes.middleware.middleware.DistanceTaskDistribution'
     - 'pro_tes.middleware.middleware.RandomTaskDistribution'
```

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Args:
+            request: Incoming request object.
+            *args: Additional positional arguments.
+            ***kwargs: Additional keyword arguments.
+
+        Returns:
+            The modified request object.
+        """
+        for main_middleware, fallback_middlewares in self.middlewares:
```
uniqueg: I wouldn't even bother with distinguishing "main" from "fallback" middlwares. Just process the entire list (if it's a list) and stop when the first one succeeds. If it's just a string, then try that. If none works, raise an error.

uniqueg (changes requested):
Please have a look at the comments. I think I would structure this differently to simplify the syntax and give an easy way to define middlewares without fallbacks. I think it would be semantically cleaner, too (and possibly also less code).

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,128 @@
...
+import importlib
+import logging
+
+from flask import current_app
+
+logger = logging.getLogger(__name__)
+
+
+class MiddlewarePipeline:
+    """Middleware pipeline."""
```
uniqueg: Can you please document args, attributes, return values and raised errors to the class, its methods and any static functions in the module?

Also, extend descriptions where necessary to facilitate understanding of the code, e.g., what is the class good for.

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(self, middlewares=None):
+        """Construct object instance."""
```
uniqueg: Please add type hints at least for method arguments and return values, ideally also for all other variables

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,128 @@
...
+        middleware_path: Middleware path in the form:
+            "module.submodule.MiddlewareClass".
+
+    Returns:
+        Middleware instance.
+    """
+    module_path, class_name = middleware_path.rsplit('.', 1)
+    module = importlib.import_module(module_path)
+    middleware_class = getattr(module, class_name)
+    return middleware_class()
```
uniqueg: Shouldn't you check whether this is a valid middleware class, i.e., whether the class inherits from the abstract middleware class (we _do_ have an abstract middleware class, right?)

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def load_middlewares_from_config(config: dict) -> list:
```
uniqueg: I think this function as well as the function below might as well be methods of the class

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def create_middleware_pipeline() -> MiddlewarePipeline:
+    """Create middleware pipeline."""
+    middleware_config = current_app.config.foca.middlewares
+    middlewares = load_middlewares_from_config(middleware_config)
+    logger.info(f"Loaded middlewares: {middlewares}")
+    pipeline = MiddlewarePipeline(middlewares)
+    logger.info(f"Created middleware pipeline: {pipeline}")
+    return pipeline
```
uniqueg: I think this code should not be in this module. Rather it should be in the module that consumes/calls the middleware class.

uniqueg (changes requested):
Still some work needed, but we're getting there

pro_tes/middleware/middleware_pipeline.py
Ayush5120: Yes you are right, will add a basic check, in the next commit ( yes we have an abstract middleware class).

pro_tes/middleware/middleware_pipeline.py
Ayush5120: yeah sure, will fix in the next commit

pro_tes/middleware/middleware_pipeline.py
Ayush5120: will fix it in the next commit.

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+        __init__(self, middlewares=None): Initializes a MiddlewarePipeline
+            instance, optionally with an initial list of middleware components.
+        add_middleware(self, middleware): Adds a middleware component to the
+            pipeline.
+        process_request(self, request, *args, **kwargs): Processes an incoming
+            request through the middleware pipeline, applying each
+            middlewares logic in sequence.
+    """
+
+    def __init__(self, middlewares=None):
```
uniqueg: Please add type hint for `middlewares`

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            initialize the pipeline. Defaults to an empty list if not provided.
+
+        Attributes:
+            middlewares (list): A list that holds the middleware components
+            in the pipeline.
+        """
+        self.middlewares = middlewares or []
+        logger.info(f"middleware: {self.middlewares}")
+
+    def add_middleware(self, middleware) -> None:
```
uniqueg: Please add type hint for `middleware`

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+            middlewares logic in sequence.
+    """
+
+    def __init__(self, middlewares=None):
+        """
+        Initialize a MiddlewarePipeline instance.
+
+        Args:
+            middlewares (list, optional): A list of middleware components to
+            initialize the pipeline. Defaults to an empty list if not provided.
```
uniqueg: Second line should be indented for args and attributes, here and in other docstrings.

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+    middleware components.
+
+    Methods:
+        __init__(self, middlewares=None): Initializes a MiddlewarePipeline
+            instance, optionally with an initial list of middleware components.
+        add_middleware(self, middleware): Adds a middleware component to the
+            pipeline.
+        process_request(self, request, *args, **kwargs): Processes an incoming
+            request through the middleware pipeline, applying each
+            middlewares logic in sequence.
```
uniqueg: This is not required. API docs create these inventories automatically.

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+
+from pro_tes.exceptions import InvalidMiddleware
+from pro_tes.middleware.middleware import AbstractMiddleware
+
+logger = logging.getLogger(__name__)
+
+
+class MiddlewarePipeline:
+    """
+    Middleware Pipeline.
```
uniqueg: No point in just repeating the name. Instead make the following description shorter to make it fit on one line.

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+
+    A class for managing and processing incoming requests through a sequence of
+    middleware components.
```
uniqueg: We can see that it's a class. Start docstrings with a verb in imperative form. E.g.:

> Apply a sequence of middlewares to requests.

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -628,3 +631,20 @@ def parse_basic_auth(auth: Optional[Dict[str, str]]) -> BasicAuth:
...
+
+    @staticmethod
+    def create_middleware_pipeline() -> MiddlewarePipeline:
+        """Create middleware pipeline.
+
+        Returns: A `MiddlewarePipeline` object that can be used to execute the
+            configured middleware on incoming task requests using the
+            `process_request` method.
+        """
+        middleware_config = current_app.config.foca.middlewares
```
uniqueg: Why create a variable here, it's just used once

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -628,3 +631,20 @@ def parse_basic_auth(auth: Optional[Dict[str, str]]) -> BasicAuth:
...
+        middlewares = MiddlewareLoader().load_middlewares_from_config(
+            middleware_config
```
uniqueg: For readability, please include the parameter name, i.e., call like this:

```py
middlewares = MiddlewareLoader().load_middlewares_from_config(
    config=current_app.config.foca.middlewares
)
```

pro_tes/middleware/middleware_pipeline.py
```diff
@@ -0,0 +1,173 @@
...
+                        **kwargs
+                    )
+                except Exception as exc:  # pylint: disable=W0703
+                    logger.exception(
+                        f"Error occurred in middleware: {exc}"
+                    )
+        return request
+
+
+class MiddlewareLoader:
```
uniqueg: I don't really see the need for a `MiddlewarePipeline` and a `MiddlewareLoader` class at this point. Both of them have just two public methods and a single attribute that's basically a list of middlewares. Why not just merge them into a single `MiddlewareHandler` class?

pro_tes/middleware/middleware.py
uniqueg, line 1: I think this module should _only_ include the abstract middleware class. _Actual_ middlewares should go in a separate package, e.g., `middlewares/`.

Then _this_ package could maybe be renamed to avoid confusion, perhaps into `middleware_handler`. I would then name _this_ module `abstract_middleware.py` and the other one `middleware_handler.py` (just like the class, see comment below).

pro_tes/ga4gh/tes/task_runs.py
```diff
@@ -628,3 +631,20 @@ def parse_basic_auth(auth: Optional[Dict[str, str]]) -> BasicAuth:
             username=auth.get("username"),
             password=auth.get("password"),
         )
+
+    @staticmethod
+    def create_middleware_pipeline() -> MiddlewarePipeline:
+        """Create middleware pipeline.
```
uniqueg: If you apply the changes as discussed below, this function will shrink to one line and one log message, and I think you don't need a wrapper function for that. Just include it in the controller method itself.

uniqueg (changes requested):
I'm really sorry to be nagging, but I really want to get this right before it's copied into proWES :-S

Ayush5120:
> I'm really sorry to be nagging, but I really want to get this right before it's copied into proWES :-S

No Problem at all, I will fix them all in the next commit.

uniqueg (approved):

PR #178, opened 2024-06-12
fix: query filter in list task endpoint

### LINKED ISSUE #175: fix: Getting list of tasks not working without name prefix
**Describe the bug**
Making a GET request to the `/tasks` endpoint proTES with an empty body always returns an empty list of tasks, even when a task has been created. This is because of [this](https://github.com/elixir-cloud-aai/proTES/blob/dev/pro_tes/ga4gh/tes/task_runs.py#L282) line in `task_runs.py`. Without a name prefix, `str(kwargs.get("name_prefix"))` returns the string `"None"` meaning `name_prefix is not None` is always True. As a result, a filter of `"^None"` is applied when querying the db.

**To Reproduce**
1. Run proTES
2. Create a task (POST to /tasks with valid body (e.g. `{"executors": [{"image": "alpine", "command": ["echo", "hello"]}]}`)
3. Get list of tasks (GET to /tasks with empty body)
4. Task list is empty

**Expected behavior**
The task list should contain the newly created task ID and status.

**Screenshots**
POST request made successfully:
Screenshot 2024-05-20 at 2 15 27 PM

GET request returns empty task list:
Screenshot 2024-05-20 at 2 16 01 PM

### LINKED ISSUE #177: fix: Handle None Values Correctly in List Task Endpoint Query Filter
When building the query filter with potential None values for name_prefix and user_id, the filter should be adjusted accordingly to avoid incorrect queries.

Expected Behavior:
- If name_prefix is None, filter_dict should be empty ({}).
- If user_id is None, it should not be added to the filter.

Actual Behavior:
Currently, if name_prefix is None, the filter becomes:
`filter_dict = {'task_original.name': {'$regex': '^None'}}`

while in the case of None Type the filter_dict should be empty like` filter_dict = {}`
The actual issue is here:
```
        name_prefix: str = str(kwargs.get("name_prefix"))

        if name_prefix is not None:
            filter_dict["task_original.name"] = {"$regex": f"^{name_prefix}"}
```
and,
`filter_dict["user_id"] = kwargs.get("user_id")`

Ayush5120:
Fixes #177
Fixes #175 

## Summary by Sourcery

This pull request addresses a bug in the list tasks endpoint by ensuring that 'user_id' and 'name_prefix' filters are only added to the filter dictionary if they are not None.

* **Bug Fixes**:
    - Fixed an issue where the 'user_id' and 'name_prefix' filters in the list tasks endpoint were not properly handling None values.

### DIFF
--- a/pro_tes/utils/db.py
+++ b/pro_tes/utils/db.py
@@ -105,9 +105,7 @@ class DbDocumentConnector:
         if projection is None:
             projection = {"_id": False}
         document_unvalidated = self.collection.find_one_and_update(
-            {
-                "worker_id": self.worker_id
-            },
+            {"worker_id": self.worker_id},
             {
                 "$set": {
                     ".".join([root, key]): value for (key, value) in
--- a/pro_tes/utils/db.py
+++ b/pro_tes/utils/db.py
@@ -2,9 +2,7 @@
 
 import logging
 from typing import Mapping, Optional
-from pymongo.collection import ReturnDocument  # type: ignore
-from pymongo import collection as Collection  # type: ignore
-
+from pymongo.collection import ReturnDocument, Collection
 from pro_tes.ga4gh.tes.models import DbDocument, TesState
 
 logger = logging.getLogger(__name__)
@@ -23,17 +21,17 @@ class DbDocumentConnector:
     """
 
     def __init__(
-            self,
-            collection: Collection,  # type: ignore
-            worker_id: str,
+        self,
+        collection: Collection,
+        worker_id: str,
     ) -> None:
         """Construct object instance."""
-        self.collection: Collection = collection  # type: ignore
+        self.collection: Collection = collection
         self.worker_id: str = worker_id
 
     def get_document(
-            self,
-            projection: Optional[Mapping] = None,
+        self,
+        projection: Optional[Mapping] = None,
     ) -> DbDocument:
         """Get document associated with task.
 
@@ -51,12 +49,12 @@ class DbDocumentConnector:
         """
         if projection is None:
             projection = {"_id": False}
-        document_unvalidated = self.collection.find_one(  # type: ignore
+        document_unvalidated = self.collection.find_one(
             filter={"worker_id": self.worker_id},
             projection=projection,
         )
         try:
-            document: DbDocument = DbDocument(**document_unvalidated)
+            document: DbDocument = DbDocument(**(document_unvalidated or {}))
         except Exception as exc:
             raise ValueError(
                 "Database document does not conform to schema: "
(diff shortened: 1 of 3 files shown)


uniqueg (commented):
Looks good to me, but before merging: @athith-g - can you please confirm that this PR solves #175 for you?

athith-g (approved):
It does fix #175 👍

uniqueg (changes requested):
That's good to hear. However, linting and tests fail - please check, @Ayush5120 :pray:

pro_tes/utils/db.py
```diff
@@ -23,17 +23,17 @@ class DbDocumentConnector:
     """
 
     def __init__(
-        self,
-        collection: Collection,
-        worker_id: str,
+            self,
+            collection: Collection,  # type: ignore
+            worker_id: str,
```
uniqueg: Why the double indenting (8 chars) on these? Did `black` do that? We should have only one level of indentation for the function signatures

pro_tes/utils/db.py
```diff
@@ -23,17 +23,17 @@ class DbDocumentConnector:
...
     ) -> None:
         """Construct object instance."""
-        self.collection: Collection = collection
+        self.collection: Collection = collection  # type: ignore
```
uniqueg: Why the `type: ignore`?

pro_tes/utils/db.py
```diff
@@ -51,7 +51,7 @@ def get_document(
         """
         if projection is None:
             projection = {"_id": False}
-        document_unvalidated = self.collection.find_one(
+        document_unvalidated = self.collection.find_one(  # type: ignore
```
uniqueg: And here?

uniqueg (commented):
Can you please explain why we are ignoring types?

pro_tes/utils/db.py
Ayush5120: my pycharm by default reformats to 8 chars indent, let me fix that in the next commit.

pro_tes/utils/db.py
Ayush5120: After removing the deprecated insert, this type: ignore is now not needed. I forgot to remove it, fixed it in the next commit

pro_tes/utils/db.py
```diff
@@ -106,17 +104,18 @@ def upsert_fields_in_root_object(
...
+                {"worker_id": self.worker_id},
+                {
+                    "$set": {
+                        ".".join([root, key]): value
+                        for (key, value) in kwargs.items()
+                    }
+                },
+                projection=projection,
+                return_document=ReturnDocument.AFTER,
+            ))
```
uniqueg: Was this formatted by `black`? If not, I think the second closing parenthesis should be on a separate line with less indent.

uniqueg (changes requested):

pro_tes/utils/db.py
Ayush5120: Fixed the indent

uniqueg (approved):

PR #181, opened 2025-01-07
build: address security vulnerabilty

lvarin:
**IMPORTANT: Please create an issue before filing a pull request! Changes need to be discussed before proceeding. Please read the [contribution guidelines](CONTRIBUTING.md).**

**Details**

Please provide enough information so that others can review your pull request. Give a brief summary of the motivation. Refer to the corresponding issue/s with `#XXXX` for more information.

**Testing**

Write the appropriate unit and integration tests, if applicable. Make sure these and all other tests pass.

**Documentation**

Please document your changes and test cases in the appropriate places, if applicable.

**Style**

Make sure your changes adhere to the coding/documentation style used throughout the project.

**Closing issues**

If your changes fix any issue/s, put `closes #XXXX` in your comment to auto-close it/them.

**Credit**

Add your credentials to the [list of contributors](contributors.md) once your pull request was merged.

## Summary by Sourcery

Build:
- Upgrade gunicorn dependency to version 22 in requirements.txt to address a security vulnerability.

PR #184, opened 2025-02-22
build: upgrade Gunicorn

### LINKED ISSUE #174: build: upgrade gunicorn
See security vulnerability here: https://github.com/advisories/GHSA-w3h3-4rj7-4ph4

Nidhi091999:
resolve #174

## Summary by Sourcery

Chores:
- Upgrade Gunicorn dependency to version 22.0.0 or higher.


uniqueg:
Thanks a lot @Nidhi091999: Do you want to try and address the linting and test issues (as well as the conflict)? I know it's not your changes that causes those, but I think the required changes are minor. You can do them in this PR or - even better - in another PR (to then be merged before this).

You can have a look at https://github.com/elixir-cloud-aai/cwl-WES/pull/258 to get some inspiration on how to fix the test issues, I think the problem is the same (with `docker compose` vs. `docker-compose` - but it might require a couple more dependency upgrades).

uniqueg:
Already addressed by #181 (Gunicorn upgrade) and #185 (for failing linting and tests)

PR #196, opened 2026-01-24
feat: add middleware management API specs

keshxvdayal:
### Detailed Implementation Plan

1. **Create OpenAPI Specification Document**
   - Create `data/specs/middleware_management.yaml`
   - Define base paths and versioning (`/ga4gh/tes/v1/middlewares`)
   - Document all endpoints:
     - `GET /middlewares` - List all configured middlewares
     - `POST /middlewares` - Add new middleware (local or from GitHub)
     - `GET /middlewares/{middleware_id}` - Get middleware details
     - `PUT /middlewares/{middleware_id}` - Update middleware configuration
     - `DELETE /middlewares/{middleware_id}` - Remove middleware
     - `PUT /middlewares/reorder` - Reorder middleware stack
     - `POST /middlewares/validate` - Validate middleware code without adding
   - Define comprehensive schemas:
     - `MiddlewareConfig` - Full middleware configuration object
     - `MiddlewareCreate` - Request body for creating middleware
     - `MiddlewareUpdate` - Request body for updating middleware
     - `MiddlewareList` - Array response with pagination support
     - `MiddlewareOrder` - Order configuration for reordering
     - `ValidationRequest` - Code validation request
     - `ErrorResponse` - Standard error format
   - Document query parameters (pagination, filtering, sorting)
   - Define all possible HTTP status codes and error responses

2. **Integrate with FOCA Configuration** 
   - Update `pro_tes/config.yaml` to reference the new spec
   - Add middleware API paths to FOCA routing configuration
   - Configure CORS and authentication settings (if applicable)
   - Set up request/response validation middleware via FOCA

3. **Write API Documentation** 
   - Create `docs/api/middleware_management.md`
   - Provide usage examples with curl commands
   - Document authentication requirements
   - Explain order/priority system for middleware execution
   - Document fallback group behavior
   - Add sequence diagrams for complex operations (reordering, fallback)

4. **Validation and Review** 
   - Validate OpenAPI spec with `openapi-generator validate`
   - Generate API documentation preview with Redoc/Swagger UI
   - Create examples for each endpoint operation
   - Peer review with team focusing on:
     - API design best practices
     - Consistency with existing proTES APIs
     - GA4GH TES specification alignment

## Summary by Sourcery

Add an OpenAPI specification and configuration wiring for a new runtime middleware management API in proTES.

New Features:
- Introduce a Middleware Management REST API specification with endpoints for listing, creating, updating, deleting, reordering, and validating middlewares.

Enhancements:
- Extend FOCA configuration to load the new middleware management OpenAPI spec with strict request/response validation and routing to middleware controllers.
- Add a MongoDB index on middleware name to support efficient lookup and management operations.

Documentation:
- Add middleware management architecture and API design documentation describing endpoints, data models, design decisions, and future work.


pro_tes/config.yaml
```diff
@@ -47,6 +47,10 @@ db:
           indexes:
             - keys:
                 id: 1
+        middlewares:
+          indexes:
+            - keys:
+                name: 1
```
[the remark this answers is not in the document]
uniqueg: Please address.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,672 @@
...
+        github_url:
+          type: string
+          description: GitHub URL for fetching middleware code
+          nullable: true
+          pattern: '^https://raw\.githubusercontent\.com/.+\.py$'
+          example: "https://raw.githubusercontent.com/user/repo/main/middleware.py"
+        code:
+          type: string
+          description: Raw Python code to validate (alternative to class_path)
+          nullable: true
```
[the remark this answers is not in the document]
uniqueg: Please address.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,672 @@
...
+          description: Execution order (omit to append to end)
+          minimum: 0
+          nullable: true
+          example: 0
+        github_url:
+          type: string
+          description: GitHub URL for fetching middleware code
+          nullable: true
+          pattern: '^https://raw\.githubusercontent\.com/.+\.py$'
+          example: "https://raw.githubusercontent.com/user/repo/main/middleware.py"
```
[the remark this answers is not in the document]
uniqueg: This is an important point, but we won't address this now, as all users are currently known. @keshxvdayal Please create a ticket for sandboxing middlewares and then close this conversation.

docs/middleware.md
uniqueg, line 1: Please write this document with a developer/admin audience in mind, and write it with respect to the current status once this PR is merged.

For example: Developers/admins who want to develop against the service or set it up won't care about a subtask. They won't to know what works and what doesn't, _how_ it works and for what use cases. And they don't want to know what _will_ be implemented, but what is currently in place.

This is the point of continuous documentation - it should ensure that the docs always represent the _current_ status, neither a previous, nor a future one. If something isn't in place yet, don't describe it - but describe it with the PR that makes it available.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,662 @@
...
+  version: 1.0.0
+  contact:
+    name: ELIXIR Cloud & AAI
+    url: https://github.com/elixir-cloud-aai/proTES
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: /ga4gh/tes/v1
```
uniqueg: consider hosting these elsewhere, given that these are not GA4GH (TES) endpoints. Just `/protes/v1` would be fine, I guess.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,662 @@
...
+            maximum: 100
+            default: 50
+        - name: offset
+          in: query
+          description: Number of results to skip (for pagination)
+          required: false
+          schema:
+            type: integer
+            minimum: 0
+            default: 0
```
uniqueg: Please double check to ensure that pagination is in line with the GA4GH pagination guide that was recently published: https://github.com/ga4gh/TASC/blob/main/recommendations/API%20pagination%20guide.md

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,662 @@
...
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/MiddlewareCreate'
+            examples:
+              local_middleware:
+                summary: Add local middleware
+                value:
+                  name: "Distance-based Router"
+                  class_path: "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
```
uniqueg: Note that this only works if the package containing the class is installed. If middlewares are shipped within proTES (which is a bit of an antipattern and will probably not be supported in the future), this will be fine - but if not, I think we need to come up with something better.

Possibly we need more than one parameter, (1) a package root path (containing a `setup.py` or `pyproject.toml` file that will allow installation) OR package name (to be installed from PyPI or another registry) OR Git repo AND (2) a class path, as an entry point to ensure that we know what code to execute in the package once it has been installed. This also makes the use of different sources (local, Git, package registry) consistent.

pro_tes/config.yaml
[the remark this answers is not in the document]
keshxvdayal: @uniqueg i have already implemented and pushed this suggestion [here](https://github.com/elixir-cloud-aai/proTES/pull/196/changes/5025c12e7afb3e332082b465b81bd51315f0e3b2#diff-2e2b0cef59a054a44d89f4d8f52647832e1e0bd02c386fdc8ae9e34763874565), please do check it. thanks a lot!

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,662 @@
...
+                  class_path: "CustomMiddleware"
+                  github_url: "https://raw.githubusercontent.com/user/repo/main/middleware.py"
+                  enabled: true
+              fallback_group:
+                summary: Add fallback group
+                value:
+                  name: "Load Balancing Group"
+                  class_path:
+                    - "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
+                    - "pro_tes.plugins.middlewares.task_distribution.random.TaskDistributionRandom"
```
uniqueg: What about combinations of local, GitHub etc. middlewares. Not all middlewares need to come from the same source, so how is the mapping done? I think the structuring of the model needs to be different - each one will need a URI (local path, GitHub URL, package registry), an entry point (class path) and possibly a type (local, Git, package registry). See comment above for more details.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,662 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MiddlewareCreateResponse'
+        '400':
+          description: Invalid request (duplicate name/class_path, invalid code)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
```
uniqueg, line 207: I think all your endpoint ops should define 400, 401, 403 and (where applicable) 404 errors.

uniqueg (changes requested):
I didn't have a chance to review everything yet, but please go ahead and address the existing comments.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+
+**MiddlewareCreateResponse**: Response after successful creation including the middleware ID, assigned order, and success message.
+
+**MiddlewareOrder**: Request body for reordering containing an array of middleware IDs in desired execution order.
+
+**ValidationRequest**: Code validation request containing package source information and entry point to validate.
+
+**ValidationResponse**: Validation result including validity boolean, validation messages, error details with line numbers, and warnings.
+
+**ErrorResponse**: Standard error response with HTTP status code, error message, and optional details.
```
uniqueg: You can replace all of this with just a reference to the Swagger Editor, with something like this: https://editor.swagger.io/?url=https://raw.githubusercontent.com/elixir-cloud-aai/proTES/refs/heads/feature/middleware-api-spec/pro_tes/api/middleware_management.yaml

The documentation is much more comprehensive, easier to parse and play with, and always up to date. Obviously you need to adapt the link to point to the path it will have once the PR is merged, not the one from _this_ branch.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Key Features
+
+**MongoDB ObjectId Format**: Uses 24-character hexadecimal strings for middleware identification. This aligns with the existing proTES database schema and provides guaranteed uniqueness.
+
+**Order-Based Execution**: Middlewares execute in ascending order. Lower order values run first. This provides clear, predictable execution flow that's easy to understand and debug.
+
+**Fallback Group Support**: Allows grouping multiple middleware sources in a single middleware entry. If the first middleware fails, the system automatically tries the next one in the list. Each middleware in a fallback group specifies its own source, package path, and entry point.
```
uniqueg: Clarify: Not the next one in the last, but the next one in the fallback stack/group.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Soft Delete Default**: DELETE operations disable rather than remove middlewares by default. This preserves execution history and allows easy rollback. Hard delete requires explicit force parameter.
```
uniqueg: This is an interesting idea, but I'm not sure this implementation although it raises some questions.

First of all, it breaks with REST semantics. `DELETE` operations delete a resource. If you want to modify a resource, you should use `UPDATE` (or possibly `PUT`), e.g., on a `enabled` parameter in the data model. This would also make it easy to reactivate it (no new endpoint needed).

Also, can this be applied to an entire fallback group at once?

And how do you actually roll back the entire state - if reproducibility is your problem here?

In any case, how do you ensure reproducibility in another proTES instance?

To be honest, there are so many questions around this that I think this should be disabled for now. Rather, you should think about how you can capture enough information about the middleware stack (put a whole copy in it, including versions) in the task logs that it can be recreated in another instance (or the same one in the future).

Speaking of versions: How can different versions of middlewares be handled? Say I want to update a repo from Git to a new commit/tag? Or a new PyPI version? And how are multiple versions handled? Do you plan to sandbox/isolate the installation artifacts or dependencies for each middleware (you absolutely should).

This brings me to another question? How does `DELETE` work? Does it just remove the middleware from the stack, or does it also undo any installations? Because over time, conflicts may accumulate, if isolation is not in place. And even if it is, this grows over time...

Of course, most of these points are discussion points for the _next_ PR. For the API model in this PR, I would just remove any provisioning for soft deletion.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+
+**MiddlewareOrder**: Request body for reordering containing an array of middleware IDs in desired execution order.
+
+**ValidationRequest**: Code validation request containing package source information and entry point to validate.
+
+**ValidationResponse**: Validation result including validity boolean, validation messages, error details with line numbers, and warnings.
+
+**ErrorResponse**: Standard error response with HTTP status code, error message, and optional details.
+
+### Key Features
```
uniqueg: None of these are currently available, so this is something that should make the docs only once these features are actually implemented. You really need to understand and internalize the continuous documentation philosophy. It doesn't matter if you think it's there next week, tomorrow or even in 2 hours. Each PR should update the documentation to represent the state of that code change - you never know what will happen.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+**Order-Based Execution**: Middlewares execute in ascending order. Lower order values run first. This provides clear, predictable execution flow that's easy to understand and debug.
+
+**Fallback Group Support**: Allows grouping multiple middleware sources in a single middleware entry. If the first middleware fails, the system automatically tries the next one in the list. Each middleware in a fallback group specifies its own source, package path, and entry point.
+
+**Soft Delete Default**: DELETE operations disable rather than remove middlewares by default. This preserves execution history and allows easy rollback. Hard delete requires explicit force parameter.
+
+**Immutable Package Configuration**: Once created, a middleware's package source and entry point cannot be changed. This prevents security risks from code substitution attacks. To change implementation, users must delete and recreate.
+
+**Multiple Package Sources**: Supports loading middleware from:
+  - **Local packages**: Installed Python packages with a class path entry point
```
uniqueg: This is actually discouraged and should be deprecated, to be removed in the future (once the middlewares inside proTES are migrated to a new repo). The reason being that it's difficult/impossible to reproduce. Plus, in most cases, users won't have access to the running instance's file system, so it's really just useful for developers and local deployments.

For now, please note that it's for development purposes and is deprecated, and list it as the third, not the first option.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+  - **GitHub repositories**: Direct Git repository URLs with setup.py or pyproject.toml
+  - **PyPI packages**: Public or private package registries with specified entry points
+
+**Source Tracking**: Records whether middleware originated from local packages, GitHub, or PyPI. Helps administrators understand deployment composition and troubleshoot issues.
+
+**Validation Endpoint**: Separate endpoint for validating middleware code before creation. Prevents deployment of broken middleware and provides immediate feedback on implementation issues.
+
+**GA4GH-Compliant Pagination**: Implements page-based pagination following the GA4GH API pagination guide with `page` and `page_size` parameters, supporting predictable result navigation.
```
uniqueg: Add a link to the pagination guide.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+pro_tes/
+├── api/
+│   └── middleware_management.yaml          (OpenAPI specification)
+└── config.yaml                             (FOCA integration)
+
+docs/
+└── middleware.md                           (This documentation)
+```
+
+## Usage Examples
```
uniqueg: Are these currently functioning, at least with stubs? Or would these API calls necessarily lead to errors? If the latter, please remove this section until they work. Reasoning is the same as above: continuous documentation

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+    "order": 0,
+    "enabled": true
+  }'
+```
+
+## Security Considerations
+
+The API includes comprehensive security controls:
+
+**Authentication Required**: All middleware management endpoints require authentication through the existing proTES security scheme.
```
uniqueg: Really? I think this should be configurable. For now we don't have authN implemented in the dashboard, so until this is the case, we wouldn't be able to use the dashboard to test it. We can discuss about enforcing this later on, but at this point, I think we are putting the cart before the horse.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+**External**:
+- OpenAPI 3.0 specification format
+- FOCA framework (Flask-based configuration)
+- Connexion (OpenAPI request routing)
+- MongoDB (persistence layer)
+
+**Internal**:
+- Existing proTES API structure
+- Current middleware plugin architecture
+- MongoDB database configuration
```
uniqueg: I don't think this is necessary at this point - there are no plans to externalize this code and run it outside of proTES.

docs/middleware.md
```diff
@@ -0,0 +1,208 @@
...
+
+## API Specification Location
+
+The complete OpenAPI 3.0 specification is available at: `pro_tes/api/middleware_management.yaml`
```
uniqueg: Group this together with the API reference to the Swagger Editor above.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+          schema:
+            type: string
+            enum: [order, name, created_at, updated_at]
+            default: order
+        - name: enabled
+          in: query
+          description: Filter by enabled status
+          required: false
+          schema:
+            type: boolean
```
uniqueg: Let's remove this for now, see comment above. Please also remove it everywhere else (models, examples, descriptions etc.).

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+    delete:
+      summary: Remove a middleware
+      description: |
+        Remove a middleware from the execution stack. By default performs soft delete
+        (sets enabled=false). Use force=true query parameter for hard deletion.
```
uniqueg: Update to remove soft deleting.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+          schema:
+            type: string
+            pattern: '^[a-f0-9]{24}$'
+        - name: force
+          in: query
+          description: Perform hard delete (permanently remove)
+          required: false
+          schema:
+            type: boolean
+            default: false
```
uniqueg: Please remove for now.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+        - Local packages: Installed Python packages with a class path entry point
+        - GitHub repositories: Git repositories containing setup.py or pyproject.toml
+        - PyPI packages: Packages from PyPI or other package registries
+        
+        If order is not specified, the middleware is appended to the end of the stack. 
+        If order is specified, existing middlewares at that position or higher are 
+        shifted up by one.
+        
+        Fallback groups can be created by providing an array of source configurations,
+        allowing mixed sources (local, GitHub, PyPI) in a single middleware entry.
```
uniqueg: How can we add a middleware to a fallback group? Do we really have to delete the entire fallback group and then reinsert? Or does this work via `PUT`? If not, I think you will need to rethink this - we should be able to insert a specific middleware at a desired position inside an existing fallback group, including if there is just one middleware available at that position. Likewise, middlewares within fallback groups should also be able to be deleted individually.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /middlewares/validate:
+    post:
+      summary: Validate middleware code
+      description: |
+        Validate middleware code without adding it to the stack. Performs static
+        analysis to check if the code is valid and safe. Useful for testing before
```
uniqueg: > valid _and_ safe

I will be curious to see how you will be doing that :)

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+                $ref: '#/components/schemas/ErrorResponse'
+
+components:
+  schemas:
+    MiddlewareConfig:
+      type: object
+      description: Complete middleware configuration object
+      required:
+        - _id
+        - name
```
uniqueg: Maybe shouldn't be required? You could make it nullable and use the package or repo name etc. as default if the user doesn't supply one.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+        - source
+        - order
```
uniqueg: This should also be nullable. Make it easy for people to use the API.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+        - enabled
+        - created_at
+        - updated_at
```
uniqueg: These should be set by the system, so should be nullable as well - a user shouldn't supply these.

But now I'm realizing why you require all of these - you are using the model in the request AND response?

Check out the `readOnly` and `writeOnly` types in OpenAPI. With these, you can give more fine-grained control over what you expect from the user in the request and what you need in the response. In any case, this may not completely solve the problem. If you want to always respond with a param, just implement it like that. But don't require it from the user if it's not absolutely necessary.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+              example:
+                - type: "local"
+                  entry_point: "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
+                - type: "github"
+                  repository: "https://github.com/org/fallback.git"
+                  entry_point: "fallback.RandomRouter"
+        order:
+          type: integer
+          description: Execution order (0 = first)
+          minimum: 0
```
uniqueg, line 479: Use a default!

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,806 @@
...
+    ValidationRequest:
+      type: object
+      description: Request body for validating middleware code
+      required:
+        - source
+      properties:
+        source:
+          $ref: '#/components/schemas/MiddlewareSource'
+
+    ValidationResponse:
```
uniqueg: I cannot reason about this without seeing the implementation (or implementation plan). Please remove it from this PR and readd when you implement it. Same with the Validation Request and endpoint. In fact, leave this out of the next PR, too. It's a bonus feature and does not have high priority at this point.

uniqueg (changes requested):
Still some ways to go, but we're getting there! :)

docs/middleware.md
uniqueg: Given that this PR does *not* have _any_ controllers (which is okay, it _shouldn't_ have), none of these features are currently supported and should, in principle, not be in the docs until the _next_ PR. Same goes for anything else that depends on the controllers being implemented and functional.

docs/middleware.md
uniqueg: You did not answer my question or in any other way address this issue, as far as I can tell.

docs/middleware.md
uniqueg: Not addressed?

docs/middleware.md
uniqueg: Not addressed?

docs/middleware.md
uniqueg: This means: Remove it here. Not addressed.

uniqueg (changes requested):

docs/middleware.md
keshxvdayal: Addressed. Entire "Security Considerations" section removed. All endpoint auth responses now state "when authentication is configured"...  making it optional/configurable for dashboard testing as requested.

docs/middleware.md
keshxvdayal: Addressed. "API Specification Location" section removed entirely. All specification references now consolidated in the "API Specification" section with the Swagger Editor link above.

pro_tes/api/middleware_management.yaml
uniqueg: This is still not addressed. Please completely remove `enabled` from the models, examples and descriptions etc. Let's keep it simple for now.

pro_tes/api/middleware_management.yaml
uniqueg: I am not sure this is addressed.

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,691 @@
...
+          type: string
+          description: Git repository URL (required for github type)
+          pattern: '^https://github\.com/.+\.git$'
+          example: "https://github.com/user/repo.git"
+        version:
+          type: string
+          description: Package version (optional, for pypi or github tag/branch)
+          example: "1.0.0"
+
+    MiddlewareCreate:
```
uniqueg: This model is redundant. As far as I can tell, it's entirely a subset of `MiddlewareConfig`. There are two options to avoid repeating yourself.

1. **Preferred:** Set `readOnly=True` on those properties in `MiddlewareConfig` that are only relevant to the response. This means that they will _not_ be expected by the client in the request, even if they are set to be `required` (they will then only be `required` in the response). FYI: `writeOnly` works in an analogous manner, if you want to add a property that is only relevant/required in a request, but should not be part of a response (e.g., a password). We don't need it here, so this is just FYI.
2. Use `allOf` in `MiddlewareConfig` to extend `MiddlewareCreate` with additional properties `MiddlewareConfig`.

**Whichever you choose: Please rename `MiddlewareConfig` (misleading!) to just `Middleware`.**

pro_tes/api/middleware_management.yaml
```diff
@@ -0,0 +1,691 @@
...
+          example: "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
+        package:
+          type: string
+          description: Package name (required for pypi type)
+          example: "protes-middleware-custom"
+        repository:
+          type: string
+          description: Git repository URL (required for github type)
+          pattern: '^https://github\.com/.+\.git$'
+          example: "https://github.com/user/repo.git"
```
uniqueg: I think this is still not very elegant, for the following reasons:

1. Conditional requirements are hidden in the `description` fields! This makes it hard to automatically validate the payload against the schema.
2. For `local`, `entry_point` may be sufficient, but what if the package is not installed? In that case, a file path (similar to `package` and `repository`) would be required, which also makes the overall usage more consistent: each source has a property that tells us where to find the package, and then a class path to identify the entry point within that package.

**Suggestion:**
Use `oneOf` with "package location" properties for all 3 enum values. For `local`, this would be optional, for the other two, it would be `required`.

You can try this (untested):

```yaml
components:
  schemas:
    MiddlewareSource:
      type: object
      description: Middleware package source configuration
      required:
        - type
        - entry_point
      properties:
        type:
          type: string
          description: Source type for the middleware package
          enum: [local, git, pypi]
        entry_point:
          type: string
          description: Class path entry point (e.g., 'package.module.ClassName')
          example: "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
        version:
          type: string
          description: Package version (optional, for pypi or github tag/branch)
          example: "1.0.0"
        package_location:
          type: string
          description: The location of the package (format and requirement vary by type).
      discriminator:
        propertyName: type
        mapping:
          local: '#/components/schemas/MiddlewareLocal'
          git: '#/components/schemas/MiddlewareGit'
          pypi: '#/components/schemas/MiddlewarePyPi'
      oneOf:
        - $ref: '#/components/schemas/MiddlewareLocal'
        - $ref: '#/components/schemas/MiddlewareGit'
        - $ref: '#/components/schemas/MiddlewarePyPi'

    MiddlewareLocal:
      allOf:
        - $ref: '#/components/schemas/MiddlewareSource'
        - type: object
          properties:
            package_location:
              type: string
              description: >
                Local file path to directory containing package. 
                Optional if the package is already installed in the environment.
              pattern: '^/.*'
              example: "/path/to/my/middleware/package"

    MiddlewareGit:
      allOf:
        - $ref: '#/components/schemas/MiddlewareSource'
        - type: object
          required:
            - package_location
          properties:
            package_location:
              type: string
              format: uri
              description: Git repository URL (must end in .git).
              pattern: '^https://github\.com/.+\.git$'
              example: "https://github.com/user/repo.git"

    MiddlewarePyPi:
      allOf:
        - $ref: '#/components/schemas/MiddlewareSource'
        - type: object
          required:
            - package_location
          properties:
            package_location:
              type: string
              description: The registered name of the package on PyPi.
              pattern: '^[a-zA-Z0-9\-_.]+$'
              example: "protes-middleware-custom"
```

uniqueg: Alternatively, if that doesn't work well, we can flatten this all:

```yaml
components:
  schemas:
    MiddlewareSource:
      type: object
      description: Middleware package source configuration
      required:
        - type
        - entry_point
        - package_location
      properties:
        type:
          type: string
          description: >
            Source type for the middleware package. 
            Determines how 'package_location' is interpreted.
          enum: [local, git, pypi]
        entry_point:
          type: string
          description: Class path entry point (e.g., 'package.module.ClassName')
          example: "pro_tes.plugins.middlewares.task_distribution.distance.TaskDistributionDistance"
        package_location:
          type: string
          description: >
            The location of the package. The expected format depends on the 'type' selected:
            - **local**: Absolute filesystem path.
            - **git**: HTTPS Git repository URL.
            - **pypi**: Registered PyPi package name.
          examples:
            local_path:
              summary: Local Directory Example
              value: "/path/to/my/middleware/package"
            git_url:
              summary: Git Repository Example
              value: "https://github.com/user/repo.git"
            pypi_name:
              summary: PyPi Package Example
              value: "protes-middleware-custom"
        version:
          type: string
          description: Package version (optional, for pypi or github tag/branch)
          example: "1.0.0"
```

This makes the model a whole lot less complex, but with some consequences:
- `package_location` would also be required for `local` (but we can argue that it's bad practice to rely on a custom package being preinstalled)
- Ability to make use of `pattern` and `format` for auto-validation.

uniqueg (changes requested):
Okay, this is starting to look much better now. There are four issues left, 2 previously existing ones, and 2 new ones. With those addressed, I think we are good to go.

Previously existing issues/comments:
- Completely remove `enabled` from all models, examples & descriptions etc.; we don't want to disable/enable middlewares for now
- Ensure there is a way how middlewares within a fallback group can be reordered and deleted

New issues/comments:
- The `MiddlewareCreate` model is redundant (using `readOnly` was also addressed previously, but ignored)
- `MiddlewareSource` needs some improvements

**After you have addressed an issue, please leave a message in the conversation where you briefly confirm what you did. This makes it much easier for reviewing. But don't resolve the issues, because that risks that issues are overlooked - leave resolving issues up to the people who have raised them.**

pro_tes/api/middleware_management.yaml
keshxvdayal: **Done! Implemented your preferred approach:**

- Removed MiddlewareCreate schema entirely
- Renamed MiddlewareConfig → Middleware
- Set readOnly: true on _id, created_at, and updated_at
- Changed required to [source] only (other fields optional in requests)
- POST endpoint now uses the unified Middleware schema

uniqueg (commented):
This looks good, thanks a lot. Please address the other 4 comments I have sent in the chat. Then we're ready to merge.

uniqueg (approved):
Great job, thanks a lot, @keshxvdayal

PR #199, opened 2026-06-18
feat: implement middleware management controllers

keshxvdayal:
### Detailed Implementation Plan

1. **Create Middleware Controller Module**  
   Implement `pro_tes.api.middlewares.controllers` to serve the middleware management API expected by FOCA/Connexion.

2. **Add Endpoint Handlers**  
   Implement handlers for:
   - `GET /middlewares` - List configured middlewares
   - `POST /middlewares` - Add a new middleware
   - `GET /middlewares/{middleware_id}` - Get middleware details
   - `PATCH /middlewares/{middleware_id}` - Update middleware configuration
   - `DELETE /middlewares/{middleware_id}` - Remove middleware
   - `PUT /middlewares/reorder` - Reorder the middleware stack

3. **Connect to FOCA Database Configuration**  
   Use the existing FOCA database wiring in config.yaml so the controller layer reads and writes from the `taskStore.middlewares` collection.

4. **Implement API Behavior**  
   Add the required runtime behavior for:
   - middleware creation with execution order assignment
   - partial updates for name/config fields
   - deletion by middleware ID
   - full stack reordering with ordered IDs
   - response shaping with API-friendly `id` fields

5. **Add Unit Tests**  
   Create controller tests using `mongomock` to validate:
   - create
   - list
   - get
   - update
   - delete
   - reorder

6. **Validate the Implementation**  
   Run the middleware controller test suite to confirm the controller layer works as expected.

7. **Document the Completion**  
   Add a short status note to the middleware docs so the API documentation reflects that subtask 2 is implemented and verified.

## Summary by Sourcery

Implement controller layer and routing for middleware management APIs backed by the taskStore.middlewares collection.

New Features:
- Expose CRUD and reordering endpoints for middlewares, including listing with pagination, creation with automatic ordering, retrieval by ID, partial updates, deletion, and stack reordering.

Enhancements:
- Add a middleware API package and wire the controller module into the existing OpenAPI/FOCA configuration for runtime use.

Documentation:
- Update middleware documentation with implementation status, code locations, and test command for the controller layer.

Tests:
- Add unit tests for the middleware controllers covering list, create, get, update, delete, and reorder behaviors.


uniqueg (commented):
Thanks @keshxvdayal!

A few comments:

Let's think about what we originally wanted to do here: We wanted to allow a (pro)TES client (e.g., the TES dashboard) to define the middlewares that are used for a specific TES job.

The middleware management doesn't really help us here, because we're creating a race condition:

- **Request 1:** Client sends one or more requests to set up the the desired middleware stack via the middleware controllers.
- **Request 2:** Client sends TES request, expecting the configured middlewares to be used.

But between calls (1) and (2), there may be other calls that change the middleware setup, and so it cannot be guaranteed that call (2) actually uses the desired middleware stack.

That being said, the middleware controllers may still be useful, as long as we add an extension to `/tasks` that allows setting the middleware stack directly. But then the question is: Wouldn't this make more sense within FOCA, so that all FOCA-based services benefit from middleware management? 🤔

But that brings up other problems: FOCA is hopelessly outdated. It still uses Connexion 2, which is in legacy maintenance mode, so it is not clear how much longer it will be supported. IMO, it doesn't make much sense to invest any considerable time into adding features into the current FOCA (or proTES, for that matter).

The issue is that migrating to Connexion 3 is a considerable undertaking. One of the changes is a migration from WSGI and Flask (Connexion 2) to ASGI and Starlette (Connexion 3). But Starlette comes with its own middleware engine (https://starlette.dev/middleware/), which uses standardized ASGI middlewares and is certainly better than whatever we would come up with.

I see three paths forward:

(1) We keep on working on the current proTES, but _only_ focus on what we want to achieve, i.e., setting the middleware stack via the `/tasks` request (as a custom proTES extension of that endpoint).
(2) We invest the time into upgrading FOCA to Connexion 3 and make use of Starlette's middleware engine, update proTES accordingly, and then add what's necessary to make use of it for our purpose (including rewriting existing middlewares).
(3) We ditch FOCA altogether (most of the services that depend on it are not maintained anymore) and build proTES and all middlewares from scratch using a fast/modern web service (e.g., written in Golang).

Given that Starlette is reasonably fast, I would tend towards option (2), as it represents a reasonable middlegroumd wrt to speed, time commitment, maintainability etc.

But it's quite some work (though perhaps not all that much with the help of agentic AI)!

Let's discuss this on a call.

